PR #38, opened 2019-03-16
Replace skill references from doubles to ints

kamronbatman:


kamronbatman:
This is actually proving to be a major pain in the ass.  I might have actually screwed up some of the math involved and I just realized that `Fixed` is derived from `Value` because of how the NonRacial and other values are determine.

From a low level, the math and data types will need to be cleaned up for this to be high impact. I might push this off for now in favor of more impactful changes.

PR #43, opened 2019-08-06
Updates Packets & Randomizer

kamronbatman:
- [X] Convert packets from classes to functions
- [ ] Move packets not used in the Core to Scripts (Distribution)
- [X] Remove God Client packets
- [X] Optimize RDRand by not optimizing it.
- [X] Deny 32bit support including zlib and rdrand
- [X] Remove RemoteAdmin (If you use it, too bad).
- [X] Add multi faceted maps/charts support for 7.0.13.0+ clients.

Future Iteration:
- [ ] Change to libuv (#44)

Example:
```csharp
    public static void SendDamageNew(NetState ns, Serial m, int amount)
    {
      if (ns == null)
        return;

      SpanWriter w = new SpanWriter(stackalloc byte[9]);
      w.Write((byte)0xBF); // Extended Packet ID
      w.Write((ushort)9); // Length

      w.Write(m);

      if (amount > 0xFFFF)
        amount = 0xFFFF;
      else if (amount < 0)
        amount = 0;

      w.Write((ushort)amount);

      ns.Send(w.Span);
    }
```
Usage:
```csharp
Packets.SendDamageNew(ns, m.Serial, 10);
```


Projects/Server/Network/Compression.cs
```diff
@@ -87,24 +87,16 @@ public static class Compression
 
     static Compression()
     {
+      if (!Core.Is64Bit)
+        throw new Exception("Core: 32-bit operating systems are not supported");
```
kamronbatman: Note that 32bit operating systems are _no longer supported_.

kamronbatman:
@3hmonkey, I didn't want this merged. It's broken.

PR #56, opened 2019-10-04
add support for modernuo.json

IcculusC:


Projects/Server/Configuration.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Server
+{
+  public class Configuration
+  {
+    public List<String> DataDirectories { get; set; } = new List<String>();
+
+    internal static string FilePath { get {
```
kamronbatman: Use expression body:
`internal static string FilePath => Path.Join(Core.BaseDirectory, "modernuo.json");`

Projects/Server/Configuration.cs
```diff
@@ -0,0 +1,120 @@
...
+      }
+      else
+      {
+        return Configuration.CreateConfiguration();
+      }
+    }
+
+    internal static string WriteConfiguration(Configuration config)
+    {
+      using (FileStream fs = new FileStream(Configuration.FilePath, FileMode.OpenOrCreate, FileAccess.Write, FileShare.Write))
```
kamronbatman: Use using expression:
`using FileStream fs = new FileStream(Configuration.FilePath, FileMode.OpenOrCreate, FileAccess.Write, FileShare.Write);`
This means no braces are needed. Using expressions dispose at the end of the parent scope.

PR #67, opened 2019-12-30
Adds Xoshiro256** RNG

kamronbatman:
Adds Xoshiro256** to ModernUO


Projects/Server/Utilities/SplitMix64.cs
```diff
@@ -0,0 +1,29 @@
...
+    public ulong Next() {
+      ulong z = x += 0x9e3779b97f4a7c15;
+      z = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9;
+      z = (z ^ (z >> 27)) * 0x94d049bb133111eb;
+      return z ^ (z >> 31);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void FillArray(ulong[] arr)
+    {
```
jaedan: For new code you should probably #nullable enable the whole file, then skip the null checks

Projects/Server/Utilities/SplitMix64.cs
```diff
@@ -0,0 +1,29 @@
...
+      if (arr == null)
+        return;
+
+      for(int i = 0; i < arr.Length; i++) arr[i] = Next();
```
jaedan: I don't know what the style guides are here strictly, but the rest of the code base puts a space after the for and always puts the body on a separate line.

Projects/Server/Utilities/Utility.cs
```diff
@@ -972,20 +972,20 @@ public static TimeSpan GetXMLTimeSpan(string timeSpanString, TimeSpan defaultVal
     public static int Dice(int numDice, int numSides, int bonus)
     {
       int total = 0;
+      uint sides = (uint)numSides;
```
jaedan: Before the cast do an if check that it's greater than or equal to 0, or change the API to use a uint as it should have been.

I always assume that there are lots of badly written scripts out there full of bugs calling these functions. Before, passing a negative value would have just given a give roll of 0. This cast would change it to possibly be a very large value.

Projects/Server/Utilities/SplitMix64.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Runtime.CompilerServices;
+
```
jaedan: You can probably just jam this class into the xoroshiro file. That's the only place it's used and it's small.

Projects/Server/Utilities/Random.cs
```diff
@@ -22,12 +22,11 @@ namespace Server
   public static class RandomImpl
   {
     private static readonly IRandomImpl _Random;
+    // private static readonly IRandomImpl _SecureRandom;
 
     static RandomImpl()
     {
```
jaedan: Don't you want to prefer xoroshiro here?

Projects/Server/Utilities/Random.cs
```diff
@@ -49,7 +48,10 @@ static RandomImpl()
 
   public interface IRandomImpl
   {
-    int Next(int c);
+    uint NextUInt32();
+    uint Next(uint n);
```
jaedan: I think you also need a separate crypto random that can be used for passwords and such. Maybe have two random implementations.

jaedan: Also, since the crypto random is only used for passwords and isn't in the performance path, can you just use the one in the standard library directly for the password code and then entirely eliminate the use of rdrand? That would simplify a ton

Projects/Server/Utilities/Utility.cs
kamronbatman: I changed it to a uint and then changed it back. I don't remember why, but you are right.

Projects/Server/Utilities/Random.cs
kamronbatman: I haven't gotten there yet. Technically this PR isn't done. I want to redo how secure RNG is used and test it. You are right though, and I will be doing that!

PR #68, opened 2020-01-02
Libuv Windows Fix

3HMonkey:
Fixes error that Libuv is not copied to output after rebuilding the project, so made it using POST-BUILD (Windowsmachine)


kamronbatman:
It works with publishing, not building. I wonder if we should use both. I'll look into it.

kamronbatman:
I am also thinking libuv for osx/Linux needs to be removed in favor of an installed copy. I may have that in my other PR. We should merge this after.

PR #70, opened 2020-01-06
Update README.md

3HMonkey:
References: https://timheuer.com/blog/deploy-aspnet-core-applications-using-self-contained-dotnet-core/


kamronbatman:
I am ok with this change, but it also requires modifying the profiles.
- Remove the Self-Contained profiles
- Remove <SelfContained>false</SelfContained> from the profiles that are not self contained

PR #84, opened 2020-02-13
Adds passhash.py utility script for admins

flamewave000:
The passhash.py script allows the admin to enter a username and password pair and it will give them the hash to use for the account. This is useful for manually resetting passwords and/or account names.

### DIFF
--- a/passhash.py
+++ b/passhash.py
@@ -1,0 +1,72 @@
+import hashlib
+from sys import argv
+from enum import Enum
+
+class Format(Enum):
+    DEFAULT = 0
+    RAW = 1
+    RAW_UPPER = 2
+    BINARY = 3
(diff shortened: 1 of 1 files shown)


kamronbatman:
Thank you for contributing!
Unfortunately we are planning on entirely replacing the accounting system with something that can interface against a DB (sqlite, mongodb, whatever) and has more robust security.

With a properly implemented account system, the admin should not have to interface with the acct data or hashes.

Let me know if you are interested in helping implement that!

flamewave000:
Sounds good, I was figuring it would need at least some rewriting given how lax the security was for password storing. I'll pull out the PR. I may be able to help with some of the DB stuff, would the DB also be used for storing all of the world model data?

PR #87, opened 2020-02-17
fixes [decorate and adds timestamp to world save

flamewave000:
Spawning systems such as the `[decorate` command would often use the raw `Activator.CreateInstance(type)` function with no additional parameters. This causes Activator to look for a class constructor in the type that has NO parameters. This will fail as many models do not have a parameterless constructor. Instead, those models contain exactly one constructor that takes exactly 1 optional parameter. The ActivatorUtil manages this by searching for a parameterless constructor first, if none are found, then it searches for a constructor with one optional parameter. It then performs the correct instantiation execution by passing it the `Type.Missing` value to the optional parameter.

All of the Activator references have been replaced with the ActivatorUtil, which will also make future fixes much more easy as they will all go through the util wrapper.

### DIFF
--- a/Projects/Server/Utilities/ActivatorUtil.cs
+++ b/Projects/Server/Utilities/ActivatorUtil.cs
@@ -1,0 +1,51 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Server.Utilities
+{
+  public static class ActivatorUtil
+  {
+    public static object CreateInstance(Type type)
+    {
+      try
+      {
+        if (type.GetConstructor(Type.EmptyTypes) != null)
+          return Activator.CreateInstance(type);
+        else
+        {
+          var emptyConstructor = type.GetConstructors().Single(info =>
+          {
+            var paramList = info.GetParameters();
+            return paramList.Length == 1 && paramList[0].IsOptional;
+          });
+          if (emptyConstructor != null)
+            return emptyConstructor.Invoke(new object[] { Type.Missing });
+          else
+            throw new TypeInitializationException(type.ToString(), new Exception($"There is no empty/default constructor for {type}"));
+        }
+      }
+      catch (NullReferenceException e)
+      {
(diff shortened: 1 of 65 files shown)


Projects/Server/World.cs
```diff
@@ -629,7 +629,7 @@ public static void Save(bool message, bool permitBackgroundWrite)
       SaveStrategy strategy = SaveStrategy.Acquire();
       Console.WriteLine("Core: Using {0} save strategy", strategy.Name.ToLowerInvariant());
 
-      Console.Write("World: Saving...");
+      Console.Write($"[{DateTime.Now.ToLongTimeString()}] World: Saving...");
```
kamronbatman: Should use UtcNow

flamewave000: I've got that switched to UTC. Good catch :)

Projects/Server/Utilities/ActivatorUtil.cs
```diff
@@ -1,51 +1,89 @@
...
-      if (args == null || args.Length == 0)
-        return CreateInstance(type);
       try
       {
-        return Activator.CreateInstance(type, args);
+        ConstructorInfo cctor = null;
+        if (args.All(x => x != null))
+          cctor = type.GetConstructor(args);
+        if (cctor == null)
+          cctor = type.GetConstructors().Single(info =>
```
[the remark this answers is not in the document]
flamewave000: I've actually given it a second pass to condense the code a little

Projects/Server/Utilities/ActivatorUtil.cs
```diff
@@ -0,0 +1,76 @@
...
+        throw new TypeInitializationException(type.ToString(), e);
+      }
+    }
+    public static object CreateInstance(Type type)
+    {
+      var cctor = GetConstructor(type);
+      var args = cctor.GetParameters();
+      if (args.Length == 0)
+        return cctor.Invoke(Type.EmptyTypes);
+      else
```
kamronbatman: No else needed since it's a returning if statement

PR #88, opened 2020-02-17
Container searching fixed and optimized

flamewave000:
The current `findItemByType()` would always fail if a predicate was not provided. These changes fix that logical issue and also convert the recursive depth-first function search to a breadth-first Queue. This will speed up the processing, and remove the security risk of stack overflow. If a bad actor with access to the `[dupe` command wanted to crash the server, they would just need to embed a ton of bags and then duplicate them many times, and then embed those. If they did it enough, they could have brought the server down.


Projects/Server/Items/Container.cs
```diff
@@ -1512,62 +1512,62 @@ private static Item RecurseFindItemByType(Item current, Type[] types, bool recur
...
-    private static void RecurseFindItemsByType<T>(Item current, bool recurse, List<T> list, Predicate<T> predicate)
-      where T : Item
-    {
-      if (current == null || current.Items.Count == 0)
-        return;
-
-      List<Item> items = current.Items;
-
-      for (int i = 0; i < items.Count; ++i)
+      var queue = new Queue<Container>();
```
kamronbatman: I don't like the idea of making a queue (or any data structure just for searching. This is a hot path too.
Maybe a pool of queues?

Projects/Server/Items/Container.cs
```diff
@@ -1512,62 +1512,62 @@ private static Item RecurseFindItemByType(Item current, Type[] types, bool recur
...
-        if (item is T typedItem)
-          if (predicate?.Invoke(typedItem) == true)
-            list.Add(typedItem);
-
-        if (recurse && item is Container)
-          RecurseFindItemsByType(item, true, list, predicate);
+        var container = queue.Dequeue();
+        foreach (var item in container.Items)
+        {
+          if (item is T typedItem && (predicate == null || predicate(typedItem)))
```
kamronbatman: Can be compressed to `predicate?.(typedItem) != false`

Projects/Server/Items/Container.cs
```diff
@@ -1512,62 +1512,62 @@ private static Item RecurseFindItemByType(Item current, Type[] types, bool recur
...
+      while(queue.Count > 0)
       {
-        Item item = list[i];
-
-        if (item is T typedItem)
-        {
-          if (predicate?.Invoke(typedItem) == true)
+        var container = queue.Dequeue();
+        foreach (var item in container.Items) {
+          if (item is T typedItem && (predicate == null || predicate(typedItem)))
```
kamronbatman: Same as above

Projects/Server/Items/Container.cs
```diff
@@ -1533,7 +1533,7 @@ private static Item RecurseFindItemByType(Item current, Type[] types, bool recur
         var container = queue.Dequeue();
         foreach (var item in container.Items)
         {
-          if (item is T typedItem && (predicate == null || predicate(typedItem)))
+          if (item is T typedItem && predicate?.Invoke(typedItem) != null)
```
kamronbatman: One of the checks is not null and another is not false.
Should be the same? I think is not false?

flamewave000: Huh, I'm not sure how that happened. I just copy-pasted the same line across. I'll have it fixed in the next commit. I just wrapped up on your other suggestion with the Queue Pool.

Projects/Server/Items/Container.cs
flamewave000: Ok so I've created a simple RefPool object that can be reused. It generates Disposable QueueRef objects that will automatically return themselves to the pool when disposed of. Figured it was the safest way to implement. I have it set to retain a max of 10 queues.

Projects/Server/Utilities/RefPool.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Server.Utilities
+{
+  public interface IRef : IDisposable { }
+
+  public class RefPool<TRef> where TRef : IRef
+  {
+    private ConcurrentBag<TRef> m_Queues = new ConcurrentBag<TRef>();
```
kamronbatman: Do we need a concurrent bag? I'm debating this internally since everything will be on the game thread (single threaded) and concurrent bag uses locks.
Maybe a concurrent queue since it doesn't use locks? Or a regular List Since we may not need concurrency?

flamewave000: Sorry, I tend to work in a lot concurrent environments, If ModernUO is not planning on multi-threaded execution, I'm totally fine with changing it to a simple collection type. Maybe a stack or queue? I think they both use a linked-list like structure that are both fast for add/remove.

flamewave000:
@kamronbatman I've revised the `RefPool` class and added some doc comments. I've also added some extra functionality to the constructor along with some safety checks.

I've replaced the `ConcurrentBag` with a regular `Stack` collection for holding the resources. Take a look and let me know what you think.

Projects/Server/Items/Container.cs
```diff
@@ -34,6 +36,7 @@ namespace Server.Items
 
   public class Container : Item
   {
+    private static QueuePool m_QueuePool = new QueuePool(QueueRef<Container>.Generate, preGenerateCount: 2, maxRefrenceRetention: 5);
```
kamronbatman: I wonder how we can know if these numbers are good. Thoughts?

flamewave000: Given that we are single threaded and non recursive, I doubt we actually will use more just one. The only way this would grow is if the container spins up parallel searches, or if the Queue Pool is shared with other classes that use queues while also calling into the search.

flamewave000:
@kamronbatman Thanks for approval, can you merge the changes? It's not giving me the option to do it myself.

Projects/Server/Items/Container.cs
kamronbatman: That is a good point. I'll think this over. It is possible we don't need this and we can just have a single static reference.
I like this though, we may use it somewhere else.

PR #91, opened 2020-02-19
Fix nullref exception in Notoriety, MobileNotoriety was being called …

Reetus:
Fix nullref exception in Notoriety, MobileNotoriety was being called with null master causing server crash,,,


flamewave000:
I'm not able to approve a PR but that looks good to me. You should add a reviewer to this PR.

PR #92, opened 2020-02-20
Spawner wouldn't spawn Types with optional parameters in ctor...

Reetus:
Spawner wouldn't spawn Types with optional parameters in ctor due to parameter length check

### DIFF
--- a/Projects/Scripts/Engines/Spawner/Spawner.cs
+++ b/Projects/Scripts/Engines/Spawner/Spawner.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Reflection;
 using Server.Commands;
 using Server.Items;
+using Server.Utilities;
 using CPA = Server.CommandPropertyAttribute;
 
 namespace Server.Mobiles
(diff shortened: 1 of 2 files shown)


Projects/Scripts/Engines/Spawner/Spawner.cs
```diff
@@ -472,49 +473,59 @@ public bool Spawn(SpawnerEntry entry, out EntryFlags flags)
...
+                  {
+                    o = ctor.Invoke(paramValues);
+                    break;
+                  }
                 }
               }
             }
           }
 
+          for (int j = 0; j < realProps.Length; j++)
```
kamronbatman: Not nested so it can be j?

Projects/Scripts/Engines/Spawner/Spawner.cs
```diff
@@ -472,49 +473,59 @@ public bool Spawn(SpawnerEntry entry, out EntryFlags flags)
...
+          {
+            if (realProps[ j ] != null)
```
kamronbatman: While we don't have formal linting, I try to stick to no spaces between index accessors.

Projects/Server/Utilities/ActivatorUtil.cs
```diff
@@ -42,32 +54,44 @@ public static ConstructorInfo GetConstructor(Type type, params Type[] args)
...
       catch (Exception e)
       {
         throw new TypeInitializationException(type.ToString(), e);
       }
     }
-    public static object CreateInstance(Type type)
+
+    private static ConstructorInfo CheckPredicate(ConstructorInfo ci, Predicate<ConstructorInfo> predicate)
+    {
+      if (predicate == null || predicate.Invoke(ci))
```
kamronbatman: Can be compressed to `predicate?. Invoke(ci) != false`

Projects/Server/Utilities/ActivatorUtil.cs
```diff
@@ -42,32 +54,39 @@ public static ConstructorInfo GetConstructor(Type type, params Type[] args)
...
       }
       catch (Exception e)
       {
         throw new TypeInitializationException(type.ToString(), e);
       }
     }
-    public static object CreateInstance(Type type)
+
+    private static ConstructorInfo CheckPredicate(ConstructorInfo ci, Predicate<ConstructorInfo> predicate)
+    {
```
flamewave000: This function seems unnecessary to me. When ever you want to check the predicate just do the following `predicate?.Invoke(ci) == true` or for a fail check: `predicate?.Invoke(ci) != true`
The only place I see the returned object used is on line 34. But that should probably be broken down into a proper if-else statement for clarity.

Reetus:
Everyone should be happy

PR #94, opened 2020-02-21
Add ConstructibleAttribute to types missing it that had it in RunUO

Reetus:
There is a bunch of types missing 'ConstructibleAttribute' that had 'ConstructableAttribute' in RunUO


kamronbatman:
These need to be gated by AccessLevel.Developer or higher. Lots of items that can be constructed need gating too. This is because they are created via code and aren't meant to be created by spawner, add command etc.

PR #97, opened 2020-02-25
Adds a new [location command for easy display of target XYZ

flamewave000:
It was a bit annoying trying to get the XYZ location within the environment. This command will place a temporary graphic on any environment tile while also stating its XYZ position. It will also display a label on any item or mobile targeted displaying their position as well. Can be used with the `[multi` command.

Aliases: `[location`, `[loc`, `[pos`

### DIFF
--- a/Projects/Scripts/Commands/LocationCommand.cs
+++ b/Projects/Scripts/Commands/LocationCommand.cs
@@ -1,0 +1,56 @@
+using Server.Commands.Generic;
+using Server.Items;
+using Server.Targeting;
+using System.Globalization;
+using System.Linq;
+
+namespace Server.Commands
+{
+  public class LocationCommand : BaseCommand
+  {
+    public LocationCommand() : base()
+    {
+      AccessLevel = AccessLevel.Counselor;
+      Commands = new[] { "Location", "Loc", "Pos" };
+      ObjectTypes = ObjectTypes.All;
+      Supports = CommandSupport.Single | CommandSupport.Multi;
+      Usage = "Location [itemIds ...]";
+      Description = "Retrieves the positional coordinates of a target. Displaying a message above the item, mobile, or environment. Environment targets will also have a graphic temporarily displayed on the tile. This graphic can be changed to the provided list of itemIds.";
+    }
+
+    public override void Execute(CommandEventArgs e, object obj)
+    {
+      int[] graphics = new[] { 0x17AF, 0X17B0, 0X17B1, 0X17B2 };
+      if (e.Arguments.Length > 0)
+        graphics = e.Arguments
+          .Select(x =>
+          {
+            int result;
+            if (x.StartsWith("0x"))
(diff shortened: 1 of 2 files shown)


kamronbatman:
I wanna test this out when I get time. Looks like an interesting feature. I am not sure if the use case outside of the normal [get location and those variants.

flamewave000:
I haven't tried out anything other than `[get location`. The primary purpose of this one is to be more visual. When clicking and item/mobile it adds a temporary label to them showing their location, while also writing it to the chat journal. But the main part is for environment. You can click on any part of the environment and it will add a temporary particle effect to the tile along with a label on the particle effect. This is good for trying to find individual world tiles. I ran into the issue when testing my new `[telgen` where I was trying to check world tile locations that were missing teleporters.

flamewave000:
These tile effects last about 3-4 seconds (the same as the standard label decay).
screenshot_2020-02-25_12-51-27

Projects/Scripts/Commands/LocationCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+    public override void Execute(CommandEventArgs e, object obj)
+    {
+      int[] graphics = new[] { 0x17AF, 0X17B0, 0X17B1, 0X17B2 };
+      if (e.Arguments.Length > 0)
+        graphics = e.Arguments
+          .Select(x =>
+          {
+            int result;
+            if (x.StartsWith("0x"))
+              return int.TryParse(x.Substring(2), NumberStyles.HexNumber, CultureInfo.CurrentCulture, out result) ? result : (int?)null;
```
kamronbatman: You can inline the out variable.

Projects/Scripts/Commands/LocationCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+        if (obj is LandTarget || obj is StaticTarget)
+        {
+          var item = EffectItem.Create(new Point3D(point), e.Mobile.Map, EffectItem.DefaultDuration);
+          foreach (int graphic in graphics)
+            Effects.SendLocationParticles(item, graphic, 10, 50, 2023);
+          item.LabelTo(e.Mobile, label);
+        }
+        else if (obj is Mobile entity) entity.SayTo(e.Mobile, label);
+        else if (obj is Item item) item.LabelTo(e.Mobile, label);
+        AddResponse($"Location: (x:{point.X}, y:{point.Y}, z:{point.Z})");
```
kamronbatman: Use the label variable here too.

flamewave000: Should I keep the "Location: " portion at the beginning? I figured it looked more neat in the journal.

kamronbatman: Yeah
`AddResponse($"Location: {label}");`

PR #113, opened 2020-04-25
Removes int XOR in GetHashCode

### LINKED ISSUE #24: Replace GetHashCode XOR
The GetHashCode overrides that xor integer values are not actually a good idea. Not even sure if the values would be unique against the contents.

These should be replaced with prime number multiply-add.

rafaelfcsouza:
Fixes #24

Introduces new Utility method (sorry 😔) `HashArray`, which hashes an array of values.
Replaced xor of ints with this method.


kamronbatman:
Per conversation with @jaedan, we don't need it.

PR #172, opened 2020-07-04
Adds listening sockets to the configuration file

kamronbatman:
- [X] Creates a JSON converter for IPEndPoint.
- [X] Adds listeners to the configuration file.
- [X] Makes the first use prompt more friendly and includes the IP address listeners.

Screen Shot 2020-07-03 at 8 37 08 PM
Screen Shot 2020-07-03 at 8 40 53 PM


kamronbatman:
Note: TcpServer should be broken up so it is more modular. This will happen organically as the engine is built out of it.

PR #184, opened 2020-07-26
Fixes getting packet length for receiving packets

### LINKED ISSUE #182: Cannot buy from vendors, VendorBuyReply packet truncated
The client vendor buy reply packet is sent whenever a player purchases an item from an NPC vendor. However to the client nothing appears to happen once you finish buying from the vendor. I've narrowed down the issue to a packet handling problem.

The [`VendorBuyReply` ](https://github.com/modernuo/ModernUO/blob/master/Projects/Server/Network/PacketHandlers.cs#L498) packet handler requires the message size in order to read the list of items requested to be purchased:

```csharp
// PacketHandlers.cs#L500
      pvSrc.Seek(1, SeekOrigin.Begin);

      int msgSize = pvSrc.ReadUInt16();
      var vendor = World.FindMobile(pvSrc.ReadUInt32());
      var flag = pvSrc.ReadByte();

      if (vendor == null)
        return;
```
However at this point the message size has already been truncated by the main `ProcessPacket` function, which slices the packet after the final position containing the Packet Id and Packet Length:

```csharp
// PacketHandlers.cs#L350
      var packet = seq.Slice(r.Position);
      var memOwner = _memoryPool.Rent((int)packet.Length);

      packet.CopyTo(memOwner.Memory.Span);

      pump.QueueWork(ns, memOwner, handler.OnReceive);
```

The `VendorBuyReply` packet handler expects to receive *all* the packet data, however it's offset, so the msg size and serial are instead read out of the buy list data and come out all garbled. The information about the packet size doesn't seem to get preserved anywhere in the `PacketReader` to replace the `ReadUInt16` call from what I could see.

I would submit a PR however with the new packet updates coming and this not being a simple change I figured it would be best to raise it here.

I'm using a temporary workaround in `ProcessPacket`:

```csharp
// PacketHandlers.cs#L350
      if (packetId == 0x3B)
        r.Seek(0, SeekOrigin.Begin);
```

### LINKED ISSUE #183: Cannot sell to vendors, VendorSellReply packet length wrong
Related to #182, when the client attempts to sell to a vendor nothing happens. 

The [packet handler](https://github.com/modernuo/ModernUO/blob/master/Projects/Server/Network/PacketHandlers.cs#L542) attempts to use the `PacketReader.Length` property in order to verify the packet is valid. However as the backing `IMemoryOwner<byte>` block for `PacketReader` always has a length of 4096 due to `SlabMemoryPool`'s `Rent(int size)` implementation ignoring the size parameter and only returning memory blocks of the default size (4096). This causes the calculated length check to always return false making the sell request fail:

```csharp

    public static void VendorSellReply(NetState state, PacketReader pvSrc)
    {
      Serial serial = pvSrc.ReadUInt32();
      var vendor = World.FindMobile(serial);

      if (vendor == null) return;

      if (vendor.Deleted || !Utility.RangeCheck(vendor.Location, state.Mobile.Location, 10))
      {
        state.Send(new EndVendorSell(vendor));
        return;
      }

      int count = pvSrc.ReadUInt16();

      if (count < 100 && pvSrc.Length == 4 + 2 + count * 6) // pvSrc.Length is always 4096
      {
        var sellList = new List<SellItemResponse>(count);

        for (var i = 0; i < count; i++)
        {
          var item = World.FindItem(pvSrc.ReadUInt32());
          int amount = pvSrc.ReadInt16();

          if (item != null && amount > 0)
            sellList.Add(new SellItemResponse(item, amount));
        }

        if (sellList.Count > 0 && vendor is IVendor v && v.OnSellItems(state.Mobile, sellList))
          state.Send(new EndVendorSell(vendor));
      }
    }
```

A workaround can be achieved by ensuring the packet header (Id + Length) doesn't get sliced by `ProcessPacket` (see #182) and then manually reading the length to use in the check:

```csharp
    public static void VendorSellReply(NetState state, PacketReader pvSrc)
    {
      pvSrc.Seek(1, SeekOrigin.Begin); // The memory slab begins at the packet start
      int msgSize = pvSrc.ReadUInt16();
      Serial serial = pvSrc.ReadUInt32();
      var vendor = World.FindMobile(serial);

      if (vendor == null) return;

      if (vendor.Deleted || !Utility.RangeCheck(vendor.Location, state.Mobile.Location, 10))
      {
        state.Send(new EndVendorSell(vendor));
        return;
      }

      int count = pvSrc.ReadUInt16();

      if (count < 100 && msgSize == (1 + 2 + 4 + 2 + (count * 6))) // Use the packet message size, not the buffer size
      {
        var sellList = new List<SellItemResponse>(count);

        for (var i = 0; i < count; i++)
        {
          var item = World.FindItem(pvSrc.ReadUInt32());
          int amount = pvSrc.ReadInt16();

          if (item != null && amount > 0)
            sellList.Add(new SellItemResponse(item, amount));
        }

        if (sellList.Count > 0 && vendor is IVendor v && v.OnSellItems(state.Mobile, sellList))
          state.Send(new EndVendorSell(vendor));
      }
    }
```

kamronbatman:
Fixes the vendor packets and batch query packet.

Closes #182 
Closes #183

### DIFF
--- a/Projects/Server/Network/MessagePumpService.cs
+++ b/Projects/Server/Network/MessagePumpService.cs
@@ -26,7 +26,7 @@ namespace Server.Network
 {
   public interface IMessagePumpService
   {
-    void QueueWork(NetState ns, IMemoryOwner<byte> memOwner, OnPacketReceive onReceive);
+    void QueueWork(NetState ns, IMemoryOwner<byte> memOwner, int length, OnPacketReceive onReceive);
     void DoWork();
   }
 
@@ -34,9 +34,9 @@ namespace Server.Network
   {
     private readonly ConcurrentQueue<Work> m_WorkQueue = new ConcurrentQueue<Work>();
 
-    public void QueueWork(NetState ns, IMemoryOwner<byte> memOwner, OnPacketReceive onReceive)
+    public void QueueWork(NetState ns, IMemoryOwner<byte> memOwner, int length, OnPacketReceive onReceive)
     {
-      m_WorkQueue.Enqueue(new Work(ns, memOwner, onReceive));
+      m_WorkQueue.Enqueue(new Work(ns, memOwner, length, onReceive));
       Core.Set();
     }
 
@@ -48,7 +48,8 @@ namespace Server.Network
         if (!m_WorkQueue.TryDequeue(out var work))
           break;
 
-        work.OnReceive(work.State, new PacketReader(new ReadOnlySequence<byte>(work.MemoryOwner.Memory)));
+        var seq = new ReadOnlySequence<byte>(work.MemoryOwner.Memory.Slice(0, work.Length));
+        work.OnReceive(work.State, new PacketReader(seq));
         work.MemoryOwner.Dispose();
       }
     }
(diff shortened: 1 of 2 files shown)


jabinb:
:+1: just tested, looks to be all fixed!

PR #229, opened 2020-09-06
Fixes NPE from Weapon Ability & Updates [cast

kamronbatman:
- [X] Fixes NPE from weapon ability
- [X] Updates [cast so it is not case sensitive, and you don't need the word "Spell" at the end.


Projects/UOContent/Spells/Base/SpellRegistry.cs
```diff
@@ -137,9 +137,13 @@ public static Spell NewSpell(int spellID, Mobile caster, Item scroll)
 
         public static Spell NewSpell(string name, Mobile caster, Item scroll)
         {
+            name = name.Replace(" ", "");
```
kamronbatman, line 140: Probably a faster way of doing this. Gotta audit all string functions.

PR #233, opened 2020-09-07
expand buffer until large enough for input bytes

jabinb:
Fixes an `ArgumentOutOfRange` exception when the `byte[]` value is bigger than the default buffer expansion size. This will increase buffer until the value fits.


kamronbatman:
Thanks! Hopefully I can finish the network code and get to updating the serialization. This method shouldn't even exist lol

kamronbatman (approved):

PR #239, opened 2020-09-11
Add workaround for Console.ReadKey

deccer:
when Console.IsInputRedirected is true for shells like git bash

### DIFF
--- a/Projects/UOContent/Misc/AccountPrompt.cs
+++ b/Projects/UOContent/Misc/AccountPrompt.cs
@@ -10,9 +10,10 @@ namespace Server.Misc
             if (Accounts.Count == 0)
             {
                 Console.WriteLine("This server has no accounts.");
-                Console.Write("Do you want to create the owner account now? (y/n)");
+                Console.Write("Do you want to create the owner account now? (y/n): ");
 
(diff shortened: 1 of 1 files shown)


Projects/UOContent/Misc/AccountPrompt.cs
```diff
@@ -10,9 +10,10 @@ public static void Initialize()
             if (Accounts.Count == 0)
             {
                 Console.WriteLine("This server has no accounts.");
-                Console.Write("Do you want to create the owner account now? (y/n)");
+                Console.Write("Do you want to create the owner account now? (y/n): ");
 
-                if (Console.ReadKey(true).Key == ConsoleKey.Y)
+                var answer = Console.ReadLine();
```
kamronbatman, line 15: Someone will have to press enter right? I wonder if we can avoid that?

deccer: Yes. Just like when entering admin username/password. I thought its a lesser evil.

kamronbatman: Now that I have thought about it, I think it is fine.

kamronbatman (approved):

PR #248, opened 2020-09-14
Fixes braces style

kamronbatman:


deccer (approved):
Looks good to me. The awkward alignment we fix as we go, as discussed.

PR #258, opened 2020-09-19
Fixes dupe exception

### LINKED ISSUE #256: [dupe should not throw if duping an unsupported item
Some items/mobs don't have empty ctors. Should send a message back explaining that instead of crashing.

As Vorspire suggested, [dupe can use the serial ctor and then replace properties using reflection.
There is a possible edge case where a normal ctor and deserialize (which wouldn't be called) does some initialization that is lost when using the serial ctor and copying properties.

kamronbatman:
- [X] Cleans up ActivatorUtil
- [X] Fixes dupe exception
- [X] Fixes a bug in BasePotion
- [X] Fixes a few possible memory leaks
- [x] Adds relevant unit tests

Closes #256


Projects/Server.Tests/Utility/ActivatorUtilTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public int Amount;
+        public string Name;
+
+        public TestTwoParamsClass(int amount, string name)
+        {
+            Amount = amount;
+            Name = name;
+        }
+    }
```
deccer, line 21: Not sure what your policy on this is, but i try to put classes, created for the purpose of unit tests, inside these classes as private ones, otherwise 1 file per class policy should be adhered.

Projects/Server/Mobiles/Mobile.cs
```diff
@@ -5738,7 +5738,7 @@ public static Item LiftItemDupe(Item oldItem, int amount)
             Item item;
             try
             {
-                item = (Item)ActivatorUtil.CreateInstance(oldItem.GetType());
+                item = (Item)oldItem.GetType().CreateInstance();
```
deccer, line 5741: .CreateInstance<T> would be nice, that would save the (Item) cast there

Projects/Server/Regions/RegionLoader.cs
```diff
@@ -46,7 +46,7 @@ public static void LoadRegions()
                     continue;
                 }
 
-                var region = ActivatorUtil.CreateInstance(type, json, JsonConfig.DefaultOptions) as Region;
+                var region = type.CreateInstance(json, JsonConfig.DefaultOptions) as Region;
```
deccer, line 49: same here I suppose.... `CreateInstance<Region>(json, ...);`

Projects/UOContent/Engines/Plants/MiscItems/GreenThorns.cs
```diff
@@ -320,7 +320,7 @@ public static GreenThornsEffect Create(Mobile from, LandTarget land)
                 if (contains)
                 {
                     var effect =
-                        (GreenThornsEffect)ActivatorUtil.CreateInstance(taep.Effect, land.Location, from.Map, from);
+                        (GreenThornsEffect)taep.Effect.CreateInstance(land.Location, from.Map, from);
```
deccer, line 323: taep ;D typo?

Projects/UOContent/Mobiles/BaseCreature.cs
```diff
@@ -2823,13 +2823,13 @@ public void AddSpellDefense(Type type)
         public Spell GetAttackSpellRandom()
         {
             var type = m_SpellAttack.RandomElement();
-            return type == null ? null : ActivatorUtil.CreateInstance(type, this, null) as Spell;
+            return type == null ? null : type.CreateInstance(this, null) as Spell;
         }
 
         public Spell GetDefenseSpellRandom()
         {
             var type = m_SpellDefense.RandomElement();
```
deccer, line 2831: Hmm should RandomElement() not always return a valid type and never null? then next line can be reduced to `return type.CreateInstance<Spell>(...)`

deccer (dismissed):
Few things, but im not sure what your angle is on them.

PR #283, opened 2020-10-23
Update Readers/Writers for Buffers

kamronbatman:
- [X] Renames PacketReader to CircularBufferReader
- [X] Fixes bugs with CircularBufferReader
- [X] Adds CircularBufferWriter
- [X] Adds SpanWriter
- [X] Adds SpanReader

### TODO for future:
- [ ] Add Span/Circular Writer/Reader extensions for reading/writing UO specific primitives
- [ ] Remove SpanExtensions

Note: I could generalize the circular buffers to work on generic types. The underlying API would be contrived. There would only be the following:
```cs
void Write(ReadOnlySpan<T> buffer);
void Write(T value);
```
and
```cs
void Read(Span<T> buffer);
T Read();
```

This also means building the current CircularBuffers off of that which doesn't change much for our use case.


Projects/Server/Buffers/CircularBufferReader.cs
```diff
@@ -239,14 +235,14 @@ public uint ReadUInt32()
...
+                        int length = firstLength + secondLength;
 
                         // Assume no strings should be too long for the stack
                         Span<byte> bytes = stackalloc byte[length];
                         First.Slice(Position).CopyTo(bytes);
-                        Second.Slice(0, secondLength).CopyTo(bytes.Slice(First.Length));
+                        Second.Slice(0, secondLength).CopyTo(bytes.Slice(firstLength));
 
-                        Position += length;
+                        Position += length + (index >= 0 ? sizeT : 0);
```
kamronbatman, line 245: We can't just add 1 or 2 for the null value. We need to actively zero it out.
Should be changed to:
`Write((byte)0);` and `Write((ushort)0);` respectively.

Projects/Server/Buffers/CircularBufferReader.cs
```diff
@@ -255,21 +251,26 @@ public uint ReadUInt32()
...
                     span = span.Slice(0, index);
                 }
+                else
+                {
+                    index = size;
+                }
             }
 
-            Position += isFixedLength ? size : index;
+            Position += isFixedLength ? size : index + sizeT;
```
kamronbatman, line 268: Same as above

Projects/Server/Buffers/CircularBufferWriter.cs
```diff
@@ -0,0 +1,394 @@
...
+
+                    return;
+                }
+            }
+            else
+            {
+                offset = Position - _first.Length;
+                sz = 0;
+            }
+
```
kamronbatman, line 285: Add check here for `sz < src.Length` so we don't do extra work for nothing.

PR #295, opened 2020-10-30
Moving packet files around

kamronbatman:
- Makes packets extension methods
- Moves packets around


deccer (approved):
I'm more of a fan of less `static` things, in order to better test them.. Like maybe turning this into some `IPacketRegistry` which then can be implemented per UO version and not have `LoginPacket703234` and `LoginPacket600293` and obviously inject that registry in other places where needed - But that would probably also mean more work on all Scripts' `Initialize()` to consume the registry there in order to subscribe to packet events and such.

However this now looks a bit more appealing than before

deccer (commented):
Hmm I think I like it that way too. IncomingPackets OutgoingPackets are better names as well

PR #337, opened 2020-12-08
[Abandonded] Converting mobile packets

kamronbatman:
## DO NOT MERGE


Projects/Server/Mobiles/Mobile.cs
```diff
@@ -3220,38 +3142,37 @@ public virtual void ProcessDelta()
...
                 else if (sendAll)
                 {
-                    ourState.Send(new MobileAttributes(m));
+                    Span<byte> attributesPacket = stackalloc byte[OutgoingMobilePackets.MobileAttributesPacketLength];
+                    OutgoingMobilePackets.CreateMobileAttributes(ref attributesPacket, m);
+                    ourState.Send(attributesPacket);
                 }
                 else if (sendAny)
                 {
+                    Span<byte> statsPacket = stackalloc byte[OutgoingMobilePackets.MobileStatsPacketLength];
```
kamronbatman, line 3158: I have a theory that it is actually slower to reuse the same struct vs using 3 separate ones due to parallelization at the JIT/CPU level. Not really worth testing though.

Projects/Server/Mobiles/Mobile.cs
```diff
@@ -3311,14 +3232,15 @@ public virtual void ProcessDelta()
...
 
-                Packet hitsPacket = null;
                 Packet statPacketTrue = null;
                 Packet statPacketFalse = null;
                 Packet hairPacket = null;
                 Packet facialhairPacket = null;
                 Packet hbpPacket = null;
                 Packet hbyPacket = null;
 
+                Span<byte> statsPacket = stackalloc byte[OutgoingMobilePackets.MobileStatsPacketLength];
```
kamronbatman, line 3242: I could probably reuse the struct above, but I would need a way to detect if it was already written. This is fine instead.

kamronbatman: One option would be to do:
```cs
statsPacket[0] = 0;
```
Instead of this line. Which, depending on the size of the stackalloc, and optimizations, may not even matter.

PR #413, opened 2021-01-17
fix(core): Converts book packets

kamronbatman:
- [X] Splits out BaseBook
- [X] Converts book packets
- [X] Makes FixHtml faster


Projects/Server/Utilities/Utility.cs
```diff
@@ -560,31 +561,10 @@ public static string FixHtml(string str)
...
-            if (hasClose)
-            {
-                sb.Replace('>', ')');
-            }
-
-            if (hasPound)
-            {
-                sb.Replace('#', '-');
-            }
+            using var sb = new ValueStringBuilder(str, stackalloc char[Math.Min(40960, str.Length)]);
```
kamronbatman, line 564: This is a stupid hack. I have to supply a stackalloc, otherwise I can't use stackalloc invalid/replacement char arrays. It gives me a leaky context error. ugh

PR #423, opened 2021-01-20
feat(mobile): Make Mobile Paralyze and Freeze functions virtual

michaelwhelehan:
I want the ability to override the Freeze and Paralyze methods in my custom scripts to be able to call a hook and determine if the player is immune to paralyzing effects and need these methods to be virtual in order to do this.


kamronbatman:
That works for me! Thank you!

kamronbatman (approved):

PR #430, opened 2021-01-25
Make MapDefinition packets configure before Account packets

michaelwhelehan:
This is my attempt to make the MapDefinition packets configure themselves before the account packets since there is an issue with the Maps not being loaded by the time the Account packets are configured and sent as arguments to the new character creation handler.


kamronbatman:
This definitely works. I think a cleaner way would be to me over map definitions to a configuration file in Data and load it in Map.cs directly in the static ctor.

PR #431, opened 2021-01-25
fix(core): Moves map definitions to a file

kamronbatman:
- [X] Moves map definitions to a `map-definitions.json` file.
- [X] Creates a TileMatrixLoader.
- [X] Moves tile matrix and multi data configurations closer to their classes.


Projects/Server/World/Map.cs
```diff
@@ -384,6 +418,23 @@ public string Name
 
         public static int MaxLOSDistance { get; set; } = 25;
 
+        public static void RegisterMap(MapDefinition mapDefinition)
```
kamronbatman, line 421: I am seeing an infinite recursion. Might want to do aggressive inlining and see if that fixes it?

PR #518, opened 2021-02-24
Modified quick-start to correctly invoke dotnet in *Nix

karolswdev:
Current documentation incorrectly states that to run **ModernUO** you must use the ``./dotnet ModernUO.dll``.

This is untrue, since ``./dotnet`` will treat this as an instruction to execute the ``dotnet`` file in current context. This is true for the following shell types: **zsh** (Mac OS X Default), **bash** (Linux default), **sh** (Linux default on minimal distributions).

This small PR fixes this inconsistency.


kamronbatman (approved):

kamronbatman:
Thanks!

PR #535, opened 2021-03-01
Match props

nullptr-w8:
Add search match for props
[global editspawner skeleton set hue 900 str 50 match str 10


kamronbatman:
This is a great idea, thank you!

PR #536, opened 2021-03-02
Match props

nullptr-w8:
Improve string compare at now is fast, but command where doesn't compile for mobile it's can't find props str, dex, etc. 
[global editspawner skeleton  set hue 200 str 150 match str 50


Projects/UOContent/Engines/Spawners/EditSpawnCommand.cs
```diff
@@ -67,45 +68,65 @@ public override void ExecuteList(CommandEventArgs e, List<object> list)
...
+                {
+                    argSpan = argSpan.SliceToLength(argSpan.Length-len);
+                }
             }
 
             var argStr = argSpan.ToString().DefaultIfNullOrEmpty(null);
             var propsStr = props.ToString().DefaultIfNullOrEmpty(null);
+            var whereStr = findmatch.ToString().DefaultIfNullOrEmpty(null);
+            //this is fix for search like  match str 10  if mob had str 100 it's find it
+            if (whereStr != null) (whereStr += " ").ToLower();
```
kamronbatman, line 95: This doesn't actually do anything 😆. ToLower() returns a lowercase string, but you aren't using it.

Projects/UOContent/Engines/Spawners/SpawnerEntry.cs
```diff
@@ -58,11 +58,13 @@ public SpawnerEntry(BaseSpawner parent, IGenericReader reader)
         [JsonPropertyName("name")]
         public string SpawnedName { get; set; }
 
+        string _props;
         [JsonPropertyName("properties")]
-        public string Properties { get; set; }
+        public string Properties { get { return _props; } set { _props = value.ToLower() + " "; } }
```
kamronbatman, line 63: I think we should avoid this. I will think of a way around it.

Projects/UOContent/Engines/Spawners/EditSpawnCommand.cs
```diff
@@ -67,45 +68,65 @@ public override void ExecuteList(CommandEventArgs e, List<object> list)
...
+                }
+                else
+                {
+                    argSpan = argSpan.SliceToLength(argSpan.Length-len);
+                }
             }
 
             var argStr = argSpan.ToString().DefaultIfNullOrEmpty(null);
             var propsStr = props.ToString().DefaultIfNullOrEmpty(null);
+            var whereStr = findmatch.ToString().DefaultIfNullOrEmpty(null);
```
kamronbatman, line 93: Note to myself: I should make a `Span<char>` version of `DefaultIfNullOrEmpty` so that we do not need to allocate  if span is null or empty.
To do this, need to look into a way to find `whitespace` using `Span<char>`. 🤔

Projects/UOContent/Engines/Spawners/EditSpawnCommand.cs
nullptr-w8: I can put to console param like  match STR  ;d

Issue #534, opened 2021-03-01
[Cleanup] Consolidate the duplicate code in Properties, PropsGump, and PropsConfig


kamronbatman:
- Consolidate PropsConfig and everything that uses it.
- Consolidate the type checking in Properties and PropsGump
- use AggressiveInlining where possible
- Breakout the `Exception` classes in Properties.cs to their own file

PR #537, opened 2021-03-02
fix(cleanup): Consolidate the duplicate code in Properties, PropsGump, and PropsConfig 

immaciek:
https://github.com/modernuo/ModernUO/issues/534


kamronbatman:
This is great, thank you! I will go over it tonight or tomorrow.

Projects/UOContent/Commands/Generic/Implementors/ObjectConditional.cs
```diff
@@ -22,9 +21,9 @@ public ObjectConditional(Type objectType, ICondition[][] conditions)
 
         public Type Type { get; }
 
-        public bool IsItem => Type == null || Type == typeofItem || Type.IsSubclassOf(typeofItem);
+        public bool IsItem => Type == null || Type == OfItem || Type.IsSubclassOf(OfItem);
```
kamronbatman, line 24: I think it might be easier to read it if we used:
```cs
Type.IsAssignableTo(OfItem);
```

Thoughts?

Projects/UOContent/Commands/Properties.cs
```diff
@@ -130,7 +96,7 @@ public static CPA GetCPA(PropertyInfo p)
 
                     if (CIEqual(p.Name, propertyName))
                     {
-                        var attr = GetCPA(p);
+                        var attr = Types.GetCPA(p);
```
kamronbatman, line 99: Is `Types` here redundant because we have a static using? 🤔

Projects/UOContent/Gumps/Props/PropsGump.cs
```diff
@@ -76,57 +45,15 @@ public class PropertiesGump : Gump
...
-
-        public static readonly string[] BoolNames = { "True", "False" };
-        public static readonly object[] BoolValues = { true, false };
 
-        public static readonly string[] PoisonNames = { "None", "Lesser", "Regular", "Greater", "Deadly", "Lethal" };
+        private static readonly string[] BoolNames = { "True", "False" };
+        private static readonly object[] BoolValues = { true, false };
 
-        public static readonly object[] PoisonValues =
+        private static readonly string[] PoisonNames = { "None", "Lesser", "Regular", "Greater", "Deadly", "Lethal" };
```
kamronbatman, line 53: I actually don't like that these are hard coded. Let's build this from PoisonImpl.
To do that, let's add a `Configure()` (which should execute after PoisonImpl).
Get the levels from `Poison.Levels` and build the PoisonNames/PoisonValues.

With that said, should BoolNames, BoolValues, PoisonNames, and PoisonValues live in `Server.Types`?

Projects/UOContent/Gumps/Props/PropsGump.cs
```diff
@@ -551,12 +478,12 @@ private static string[] GetCustomEnumNames(Type type)
             return ce.Names;
         }
 
-        public static bool HasAttribute(Type type, Type check, bool inherit) =>
+        private static bool HasAttribute(Type type, Type check, bool inherit) =>
```
kamronbatman, line 481: Lots of these static `Type` related functions can probably go into `Server.Types` and stay public so they can be reused. This way we can create more `Props` stuff without having to recreate them.

Projects/UOContent/Commands/Properties.cs
kamronbatman: Also I am curious how this works because `GetCPA` looks like it is still in PropertiesGump? I'll open this in an IDE later.

kamronbatman:
sorry for creating a merge conflict. I fixed the name of the file and the copyright. Probably best to fix it manually, then commit (don't merge with main). Then update against main.

Projects/UOContent/Commands/Generic/Implementors/ObjectConditional.cs
immaciek: I think `OfItem.IsAssignableFrom(t)` would be marginally faster for some cases than checking for subclass. Personally I have no preference as far as readability is concerned.

immaciek: Dropped type equality and replaced with `IsAssignableTo`  in 10d50a9

Projects/UOContent/Commands/Properties.cs
immaciek: Dropped `Types.` and remove dup `GetCPA` from `Types.cs` in 87903dd

Projects/UOContent/Gumps/Props/PropsGump.cs
immaciek: Moved this to a newly created `TypesAttributes.cs` to separate `Type` constants and helpers that access attributes via reflection in acfc5ed.

Projects/UOContent/Gumps/Props/PropsGump.cs
immaciek: Moved boolean and poison consts to `Types` in acfc5ed.
Let's address `Poison`  cleanup as a separate task.

kamronbatman:
Closing this in favor of #540 since it is merged with `main`. Going to make future adjustments there.

PR #547, opened 2021-03-10
fix(spawner): Adds `match` keyword for spawner edit command

nullptr-w8:
Adds `match` keyword to filter which entries should be updated.


kamronbatman:
Thank you!

PR #558, opened 2021-03-21
[WIP] Add possess command

immaciek:
`[possess` allows GMs take control of mobs or npcs. It's a useful functionality for animating gameplay on pol implementations.

Under the hood the command stables the targeted mob and switches clothing.

TODO: 
- `[unpossess` should not be targeted
- instead of swapping items, they could be packed in a special container and stored away with the stabled mob

### DIFF
--- a/Projects/Server/Mobiles/Mobile.cs
+++ b/Projects/Server/Mobiles/Mobile.cs
@@ -1858,6 +1858,9 @@ namespace Server
             }
         }
 
+        [CommandProperty(AccessLevel.GameMaster)]
+        public Container PossessContainer { get; set; } = null;
+
         public virtual bool KeepsItemsOnDeath => m_AccessLevel > AccessLevel.Player;
 
         public bool HasTrade => m_NetState?.Trades.Count > 0;
--- a/Projects/Server/Mobiles/Mobile.cs
+++ b/Projects/Server/Mobiles/Mobile.cs
@@ -288,8 +288,7 @@ namespace Server
     public enum PossessType : byte
     {
         None,
-        Possessed,
-        Possessing
+        Possessed
     }
 
     [Flags]
@@ -672,6 +671,8 @@ namespace Server
         public virtual bool NewGuildDisplay => false;
 
         public List<Mobile> Stabled { get; private set; }
+
+        public Mobile PossessedMobile => Stabled.Find(mob => mob.PossessType == PossessType.Possessed);
 
         [CommandProperty(AccessLevel.Counselor, AccessLevel.GameMaster)]
         public VirtueInfo Virtues { get; private set; }
--- a/Projects/Server/Mobiles/Mobile.cs
+++ b/Projects/Server/Mobiles/Mobile.cs
@@ -70,6 +70,13 @@ namespace Server
         public override bool CheckCondition() => true;
     }
 
+    public class PossessionSkillMod : DefaultSkillMod
+    {
+        public PossessionSkillMod(SkillName skill, double value) : base(skill, false, value)
+        {
+        }
+    }
+
     public abstract class SkillMod
     {
         private bool m_ObeyCap;
@@ -667,6 +674,8 @@ namespace Server
         public List<Mobile> Stabled { get; private set; }
 
         public Mobile PossessedMobile => Stabled.Find(mob => mob.Possessed);
+
+        public bool IsPossessing => PossessedMobile != null;
 
         [CommandProperty(AccessLevel.Counselor, AccessLevel.GameMaster)]
         public VirtueInfo Virtues { get; private set; }
@@ -2165,6 +2174,18 @@ namespace Server
                     OnRawStatChange(StatType.Dex, oldValue);
                 }
             }
+        }
+
+        public int GetStat(StatType statType)
+        {
+            return statType switch
+            {
+                StatType.Str => Str,
+                StatType.Int => Int,
+                StatType.Dex => Dex,
+                StatType.All => Dex + Str + Int,
+                _            => 0
+            };
         }
 
         /// <summary>
--- a/Projects/Server/Mobiles/Mobile.cs
+++ b/Projects/Server/Mobiles/Mobile.cs
@@ -285,6 +285,13 @@ namespace Server
         Locked
     }
 
+    public enum PossessType : byte
+    {
+        None,
+        Possessed,
+        Possessing
+    }
+
     [Flags]
     [CustomEnum(new[] { "North", "Right", "East", "Down", "South", "Left", "West", "Up" })]
     public enum Direction : byte
@@ -546,6 +553,7 @@ namespace Server
         private DateTime m_NextWarmodeChange;
         private bool m_Paralyzed;
         private Timer m_ParaTimer;
+        private PossessType m_PossessType;
         private bool m_Player;
         private Poison m_Poison;
         private Prompt m_Prompt;
@@ -610,6 +618,8 @@ namespace Server
                 World.MobileTypes.Add(ourType);
                 TypeRef = World.MobileTypes.Count - 1;
             }
+
+            m_PossessType = PossessType.None;
         }
 
         public static bool DragEffects { get; set; } = true;
@@ -803,6 +813,12 @@ namespace Server
                     }
                 }
             }
+        }
+
+        [CommandProperty(AccessLevel.GameMaster)]
+        public PossessType PossessType {
+            get => m_PossessType;
+            set => m_PossessType = value;
         }
 
         [CommandProperty(AccessLevel.GameMaster)]
@@ -2559,7 +2575,9 @@ namespace Server
                 return;
             }
 
-            writer.Write(32); // version
+            writer.Write(33); // version
+
+            writer.Write((byte) m_PossessType);
 
             writer.WriteDeltaTime(LastStrGain);
             writer.WriteDeltaTime(LastIntGain);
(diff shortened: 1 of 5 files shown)


Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Collections.Generic;
+using Server.Commands.Generic;
+using Server.Mobiles;
+
+namespace Server.Commands
+{
+    internal static class SwapItems
+    {
+        static readonly List<Layer> SPECIAL_LAYERS = new() { Layer.Backpack, Layer.Bank, Layer.Mount };
+        public static void swapItems(Mobile target, PlayerMobile caster)
```
karolswdev, line 11: public methods should follow ```PascalCase``` format :)

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,136 @@
...
+                Properties.SetValue(e.Mobile, caster, "blessed", "false");
+
+                SwapItems.swapItems(toRelease, caster);
+
+                toRelease.MoveToWorld(caster.Location, caster.Map);
+                toRelease.Direction = caster.Direction;
+                toRelease.PossessType = PossessType.None;
+
+                BuffInfo.RemoveBuff(caster, BuffIcon.Incognito);
+                AddResponse("You've released control of " + toRelease.Name);
```
karolswdev, line 132:
```suggestion
                AddResponse($"You've released control of {toRelease.Name}");
```
👍

karolswdev (changes requested):

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,151 @@
+﻿using System;
+using System.Collections.Generic;
+using Server.Commands.Generic;
+using Server.Mobiles;
+
+namespace Server.Commands
+{
+    internal static class Swap
+    {
+        private static readonly List<Layer> SPECIAL_LAYERS = new() { Layer.Backpack, Layer.Bank, Layer.Mount };
```
kamronbatman, line 10: use `HashSet<Layer>` and naming `_specialLayers`

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        private static List<Item> GetItemsFromLayers(IEnumerable<Item> items) =>
+            new List<Item>(items).FindAll(item => !SPECIAL_LAYERS.Contains(item.Layer));
+        public static void Items(Mobile target, PlayerMobile caster)
```
kamronbatman, line 14: nit: empty line between L13 and L14

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            foreach (Item item in casterItems)
+            {
+                caster.RemoveItem(item);
+            }
+
+            foreach (Item item in targetItems)
+            {
+                caster.EquipItem(item);
+                target.RemoveItem(item);
```
kamronbatman, line 27: I don't think an item can be in two containers at once. Either you don't need to do a `RemoveItem` or it should be done before the equip.

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,151 @@
...
+using Server.Mobiles;
+
+namespace Server.Commands
+{
+    internal static class Swap
+    {
+        private static readonly List<Layer> SPECIAL_LAYERS = new() { Layer.Backpack, Layer.Bank, Layer.Mount };
+
+        private static List<Item> GetItemsFromLayers(IEnumerable<Item> items) =>
+            new List<Item>(items).FindAll(item => !SPECIAL_LAYERS.Contains(item.Layer));
```
kamronbatman, line 13: Linq is not efficient, so let's avoid using it:
```cs
private static List<Item> GetItemsOnMobile(Mobile m)
{
    var list = new List<Item>(_specialLayers.Count);
    for (var i = 0; i < _specialLayers.Count; i++)
    {
        var item = m.FindItemOnLayer(_specialLayers[i]);
        if (item != null)
        {
            list.Add(item);
        }
    }

    return list;
}
```

kamronbatman: I would probably not even make this function, avoid the list, and do everything inline

Projects/UOContent/Commands/Possess.cs
kamronbatman, line 10: Do we want to skip the mount and backpack? If we do, firing a weapon or casting a spell will not work right. And a mount is part of possessing some NPCs, right?
We might need to "lock" the player so they can't unmount and "break" the possession identity.

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,151 @@
...
+
+                target.PossessType = PossessType.Possessed;
+
+                Swap.Items(target, caster);
+                Swap.Titles(target, caster);
+
+                caster.Stabled.Add(target);
+                target.Internalize();
+
+                Properties.SetValue(e.Mobile, caster, "blessed", "true");
```
kamronbatman, line 93: We should not use `Properties`.. why can't we just do:
`caster.Blessed = true`?

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,151 @@
...
+                {
+                    AddResponse("You can't possess a player");
+                    return;
+                }
+
+                caster.BodyMod = target.Body;
+                caster.NameMod = target.Name;
+                caster.HueMod = target.Hue;
+                caster.SetHairMods(target.HairItemID, target.FacialHairItemID);
+                caster.HairHue = target.HairHue;
```
kamronbatman, line 79: We need a `hair hue mod` or a way of storing this. Same with facial hair.

Projects/Server/Mobiles/Mobile.cs
```diff
@@ -285,6 +285,12 @@ public enum StatLockType : byte
         Locked
     }
 
+    public enum PossessType : byte
```
kamronbatman, line 288: Are there other types of possession to consider? I am wondering if we should just use a bool.

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,151 @@
...
+            ObjectTypes = ObjectTypes.Mobiles;
+            Usage = "Possess";
+            Description = "Takes control of a mobile";
+        }
+
+        public override void Execute(CommandEventArgs e, object obj)
+        {
+            Console.WriteLine("Executing possess command");
+
+            if (e.Mobile is PlayerMobile caster && obj is Mobile target)
```
kamronbatman, line 60: We might want to exclude some mobile types?
* pets
* summons
* hireables?
* vendors?
* questors?
* guards?

Projects/Server/Mobiles/Mobile.cs
immaciek: Agreed, I started off with `None`, `Possessed` and `Possessing` initially but realized keeping the `Possessing` state is unnecessary.

Projects/UOContent/Commands/Possess.cs
```diff
@@ -0,0 +1,157 @@
...
+                caster.RemoveItem(item);
+            }
+
+            foreach (Item item in targetItems)
+            {
+                target.RemoveItem(item);
+
+                if (!caster.EquipItem(item))
+                {
+                    Console.WriteLine($"Nie udało się ubrać {item.Name} na {target.Name}");
```
kamronbatman, line 31: I wish we had localization. Until we do, probably need to translate this from polish.
```cs
$"Cannot equip {item.Name} on {target.Name}"
```

kamronbatman: Also a few more questions:
Should this be a failure and force a revert? Maybe we can test for equipping before doing that so we don't get into an awkward state?

Next, should we `out string error` or something like that so we can do `AddResponse` instead of `Console.WriteLine`? 🤔

Projects/UOContent/Commands/Possess.cs
kamronbatman: Even though this says it is outdated, it is still relevant.

Projects/UOContent/Commands/Possess.cs
immaciek: This approach will indeed need to change to something else. Perhaps caster's clothes need to be packed in a container and stored away somewhere they can be retrieved from.

I was using the console log to demonstrate to my colleagues how it wouldn't work on a mob such as a chicken (ie. upon unpossessing you don't get your clothes back). ;)

Sorry, will be more mindful with commits on this branch going forward... or I can close the PR for now and re-open it once it's ready for review in a few days time.

Projects/UOContent/Commands/Possess.cs
immaciek: Good idea - note this was based on incognito spell.

kamronbatman: We should add a HairHueMod and FacialHueMod for that too then! Good find!

Projects/UOContent/Commands/Possess.cs
kamronbatman: No problem, I am ok with messy commits. We squash when it gets merged in at the end. Thank you!

Projects/UOContent/Commands/Possess.cs
immaciek: np - I suggest creating a separate task for that to avoid scope creep.

immaciek:
After careful consideration I think there's no point digging further into this. 
Too many edge cases, too much faff and effort for too little benefit.

PR #571, opened 2021-04-15
fix: Configure expansion from config

ppardalj:
Expansion was not being set when reading ServerConfiguration, causing it to be always `Expansion.None` and many features to not work.


kamronbatman:
Good call! Thank you!

ppardalj:
Done :) thank you!

kamronbatman (approved):

PR #572, opened 2021-04-15
feat: Command prefix is now configurable via settings

ppardalj:
Many shard owners ask to set command prefix to other character rather than "`[`" (for example `"."`). Let's allow them to do it without changing code :)

### DIFF
--- a/Projects/UOContent/Commands/Handlers.cs
+++ b/Projects/UOContent/Commands/Handlers.cs
@@ -21,7 +21,7 @@ namespace Server.Commands
     {
         public static void Initialize()
         {
-            CommandSystem.Prefix = ServerConfiguration.GetOrUpdateSetting("command.prefix", "[");
+            CommandSystem.Prefix = ServerConfiguration.GetOrUpdateSetting("commandsystem.prefix", "[");
 
             Register("Go", AccessLevel.Counselor, Go_OnCommand);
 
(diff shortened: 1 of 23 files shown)


Projects/UOContent/Commands/Handlers.cs
```diff
@@ -21,7 +21,7 @@ public static class CommandHandlers
     {
         public static void Initialize()
         {
-            CommandSystem.Prefix = "[";
+            CommandSystem.Prefix = ServerConfiguration.GetSetting("command.prefix", "[");
```
kamronbatman, line 24: Let's use `GetOrUpdateSetting` so we can set a default in the config file so people can modify it there.

kamronbatman (changes requested):

kamronbatman:
Great job, thank you!

Projects/UOContent/Commands/Handlers.cs
kamronbatman, line 24: Also a nitpick, but I am wondering if we should make the feature `commandsystem` so the config would be `commandsystem.prefix`.

Thoughts?

ppardalj: Good idea, changing it! Thanks for feedback!

ppardalj: Sounds good :)

kamronbatman (approved):

PR #573, opened 2021-04-15
Feat/log with serilog

ppardalj:
This pull request introduces a modern logging mechanism with [Serilog](https://serilog.net/), a popular structured logging library for .NET, to replace old logs with `Console.WriteLine`.

Adopting a proper structured logging library would allow us from benefit in many aspects: consistent format/styling, centralized logging (to file, or some log aggregation stack such as ELK), to name a few.

Only some usages of `Console.WriteLine` were migrated. Migrating all usages would be extremely time consuming, so I suggest to approach it in future pull requests.


.gitignore
```diff
@@ -9,11 +9,9 @@
 /Distribution/Backups
 /Distribution/Saves
 /Distribution/docs
-/Distribution/zlib.dll
 /Distribution/libz.dylib
 /Distribution/libz.so
-/Distribution/ZLib.Bindings.dll
-/Distribution/Microsoft.Toolkit.HighPerformance.dll
+/Distribution/*.dll
```
kamronbatman, line 14: Hmm I am debating if we should do this. My biggest question is how will people consume this repo. If they fork it and start dropping files in, I don't want them to be excluded.

kamronbatman:
This is a great start! One thing I am interested in would be benchmarks. Feel free to link them, or add one to the Benchmarks project.
Mainly I want to see how much (if anything) we save from using Serilog to output to Console and/or log files over Console.WriteLine now.

Thank you!!

kamronbatman:
This seems to indicate Serilog is faster than NLog:
https://github.com/panchengtao/LoggingPerformance

kamronbatman:
Also I am wondering if we should use: https://github.com/serilog/serilog-extensions-hosting
So if we want to replace the logger, we aren't constrained by the interface/abstraction.

Thoughts?

.gitignore
ppardalj: That's a fair point but still, dll files in the build output directory should never be commited into version control, don't you think? Moreover, there's already the `.csproj` files which is a manifest of what libraries and what specific versions are needed, and that one is already checked in.

ppardalj:
> Also I am wondering if we should use: https://github.com/serilog/serilog-extensions-hosting
> So if we want to replace the logger, we aren't constrained by the interface/abstraction.
> 
> Thoughts?

As far as I understood, that package routes frameworks logs to Serilog sink, but we are not using any framework or host for ModernUO so I don't see how we could benefit.

If we ever want to change the logging library, there are 2 options that come to my mind:
- Using `ILogger` interface from `Microsoft.Extensions.Logging` (https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-5.0). This is the recommended way for example in ASP.NET Core applications or any app that use dependency inyection, however it does not play nicely with scenarios with lots of static and singletons, like ModernUO.
- Create some static wrapper classes to encapsulate the logging library (some `Server.Logger`) that we can access from everywhere, but still hides the implementation details. I probably prefer this one for this context.

Which one do you like most?

ppardalj:
> * I know Serilog can be made async, which is where we should get the biggest benefit. So I am hoping the benchmarks help show that.

I implemented a benchmark. Here are the results:

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
AMD Ryzen 5 2600, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.202
  [Host]        : .NET Core 5.0.5 (CoreCLR 5.0.521.16609, CoreFX 5.0.521.16609), X64 RyuJIT
  .NET Core 5.0 : .NET Core 5.0.5 (CoreCLR 5.0.521.16609, CoreFX 5.0.521.16609), X64 RyuJIT

Job=.NET Core 5.0  Runtime=.NET Core 5.0  InvocationCount=1  
UnrollFactor=1  

|                      Method |        Mean |        Error |      StdDev |      Median |
|---------------------------- |------------:|-------------:|------------:|------------:|
|        TestConsoleWriteLine | 5,553.62 us | 3,086.033 us | 8,653.57 us |    98.10 us |
|      TestSerilogConsoleSink | 6,758.07 us | 1,878.861 us | 5,539.86 us | 6,921.80 us |
| TestSerilogAsyncConsoleSink |    77.01 us |     6.260 us |    17.86 us |    73.95 us |

It's the first benchmark I implement so I'm not sure it's done properly, can you please review it?

```

kamronbatman:
> > Also I am wondering if we should use: https://github.com/serilog/serilog-extensions-hosting
> > So if we want to replace the logger, we aren't constrained by the interface/abstraction.
> > Thoughts?
> 
> As far as I understood, that package routes frameworks logs to Serilog sink, but we are not using any framework or host for ModernUO so I don't see how we could benefit.
> 
> If we ever want to change the logging library, there are 2 options that come to my mind:
> 
> * Using `ILogger` interface from `Microsoft.Extensions.Logging` (https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-5.0). This is the recommended way for example in ASP.NET Core applications or any app that uses dependency injection, however it does not play nicely with scenarios with lots of static and singletons, like ModernUO.
> * Creating some static wrapper class to encapsulate the underlying logging library (some `Server.Logger`) that we can access from everywhere, but still hides the implementation details. I probably prefer this one for this context.
> 
> Which one do you like most?

I like your idea of a Server.Logger! That way if someone wants to completely change the logging for their shard, they can.

Also keep in mind we may want multiple loggers eventually. One for normal (console) stuff. One for networking. One to replace commands.
Each one would probably be a singleton?

.gitignore
kamronbatman: I guess this is fine. Let's do the same for .so and .dylib files.

Projects/Server/Main.cs
```diff
@@ -359,6 +360,11 @@ public static void Main(string[] args)
             AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
             AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
 
+            Log.Logger = new LoggerConfiguration()
+                .WriteTo.Console(
+                    outputTemplate: "[{Timestamp:HH:mm:ss} {Level:u3}] {Message:lj} <s:{SourceContext}>{NewLine}{Exception}")
+                .CreateLogger();
```
immaciek, line 366: I already wrote some code that makes this configurable from a json file using a similar pattern as `ServerConfiguration.cs`, `EmailConfiguration.cs` etc.. is it ok to add it in scope for this change ?

.gitignore
ppardalj: Done!

Projects/Server/Main.cs
ppardalj: I would try to get this PR merged then move the configuration to json, what do you think?

ppardalj:
> I like your idea of a Server.Logger! That way if someone wants to completely change the logging for their shard, they can.
> 
> Also keep in mind we may want multiple loggers eventually. One for normal (console) stuff. One for networking. One to replace commands.
> Each one would probably be a singleton?

Hi @kamronbatman , I updated the PR with a patch that encapsulates all the serilog stuff behind a custom interface `Server.ILogger`, this way we can change the underlying logger library easily if we want to.

Regarding the multiple loggers, a typical pattern is to have a static logger for each class. Serilog under the hood will log the type name as the `SourceContext` property and you can do awesome magic with logging config (cc @immaciek ^^) based on this property.

Typical way you declare a logger is:

```chsarp
public partial class NetState : IComparable<NetState>
{
    private static readonly ILogger logger = LogFactory.GetLogger(typeof(NetState));
    // rest of the class...
}
```

If we go like this, we could do things such as, for example, log only network errors to file (`NetState` source context), the rest keep it only to console.

Let me know your thoughts. I hope this is closer to being merged now :)

Projects/Server/Maps/MapLoader.cs
```diff
@@ -85,21 +88,28 @@ public static void Configure()
...
-                Console.WriteLine(string.Join(Environment.NewLine, failures));
-                Utility.PopColor();
+                logger.Warning(
+                    "Map Definitions loaded with failures ({0} maps, {1} failures) ({2:F2} seconds)",
+                    count,
+                    failures.Count,
+                    stopwatch.Elapsed.TotalSeconds
+                );
+
+                if (failures.Count > 0)
```
kamronbatman, line 100: This if statement is not needed since it is inside of an if statement that already does that.

Projects/Server/Serialization/GenericPersistence.cs
```diff
@@ -73,8 +73,8 @@ void Deserialize(string savePath)
                 catch (Exception e)
                 {
                     Utility.PushColor(ConsoleColor.Red);
-                    Persistence.WriteConsoleLine($"***** Bad deserialize of {name} *****");
-                    Persistence.WriteConsoleLine(e.ToString());
+                    Console.WriteLine($"***** Bad deserialize of {name} *****");
```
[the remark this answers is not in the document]
ppardalj: Yes, I changed to log always to console because it is part of the interactive world load (i.e. right after this user is asked whether to continue loading or not). IMHO doesn't make much sense to prompt the user via console but asking her elsewhere ^^u

Projects/Server/Maps/MapLoader.cs
ppardalj: True! changing it

ppardalj:
@kamronbatman I think all the feedback has been addressed, how can we move on with this PR?

kamronbatman:
> @kamronbatman I think all the feedback has been addressed, how can we move on with this PR?

I'll take a look at it, and do some testing when I get home later today. Thank you so much for the work

kamronbatman (approved):

kamronbatman:
Thank you very much! Some cleanup work is needed, but we can do that in subsequent PRs.
My benchmarks look like this:
```
BenchmarkDotNet=v0.12.1, OS=macOS 11.2.3 (20D91) [Darwin 20.3.0]
Intel Core i9-9980HK CPU 2.40GHz, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.202
  [Host]        : .NET Core 5.0.5 (CoreCLR 5.0.521.16609, CoreFX 5.0.521.16609), X64 RyuJIT
  .NET Core 5.0 : .NET Core 5.0.5 (CoreCLR 5.0.521.16609, CoreFX 5.0.521.16609), X64 RyuJIT

Job=.NET Core 5.0  Runtime=.NET Core 5.0

|                      Method |       Mean |     Error |     StdDev |
|---------------------------- |-----------:|----------:|-----------:|
|        TestConsoleWriteLine | 124.429 ms | 2.6658 ms |  7.7763 ms |
|      TestSerilogConsoleSink | 140.609 ms | 3.7365 ms | 10.9585 ms |
| TestSerilogAsyncConsoleSink |   4.278 ms | 0.0837 ms |  0.1028 ms |

// * Hints *
Outliers
  BenchmarkConsoleLogging.TestConsoleWriteLine: .NET Core 5.0        -> 2 outliers were removed, 3 outliers were detected (103.70 ms, 148.18 ms, 152.53 ms)
  BenchmarkConsoleLogging.TestSerilogConsoleSink: .NET Core 5.0      -> 1 outlier  was  removed (180.58 ms)
  BenchmarkConsoleLogging.TestSerilogAsyncConsoleSink: .NET Core 5.0 -> 1 outlier  was  detected (4.03 ms)
  ```

PR #579, opened 2021-04-19
feat(mobile): Allow Criminal and ExpireCriminalTimer properties to be…

michaelwhelehan:
… overridden

This change allows the Criminal and CriminalTimer
properties to be overridden in derived classes ie:
PlayerMobile.

We have a use-case where we need to access the
ExpireCriminalTimer property and override Criminal
to display the data in a custom CUO status gump.

### DIFF
--- a/Projects/Server/Mobiles/Mobile.cs
+++ b/Projects/Server/Mobiles/Mobile.cs
@@ -494,7 +494,6 @@ namespace Server
 
         private Timer m_ExpireAggrTimer;
         private Timer m_ExpireCombatant;
-        private Timer m_ExpireCriminal;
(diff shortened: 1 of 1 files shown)


kamronbatman (approved):

Projects/Server/Mobiles/Mobile.cs
```diff
@@ -1889,8 +1888,10 @@ public int ShortTermMurders
             }
         }
 
+        public virtual Timer ExpireCriminalTimer { get; set; }
```
kamronbatman, line 1891: Let's not make this virtual. Then it is ready to go!

kamronbatman (changes requested):

PR #585, opened 2021-04-25
fix(core): Fixes displaying some doubles

kamronbatman:
Note:
- I am not going to go through the entire code base and fix rounding errors for doubles. It is true that most of the variables _should have been integers anyway_ and then divided into the appropriate value w/ truncation when displayed. I'll leave that up to others to make PRs to fix if they feel the need.

### DIFF
--- a/Projects/UOContent/Items/Special/Special Scrolls/ScrollofTranscendence.cs
+++ b/Projects/UOContent/Items/Special/Special Scrolls/ScrollofTranscendence.cs
@@ -36,15 +36,7 @@ namespace Server.Items
         public override void GetProperties(ObjectPropertyList list)
         {
             base.GetProperties(list);
-
-            if (Value == 1)
-            {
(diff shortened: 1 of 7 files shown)


Projects/UOContent/Items/Special/Special Scrolls/ScrollofTranscendence.cs
```diff
@@ -36,15 +36,7 @@ public ScrollofTranscendence(Serial serial) : base(serial)
...
-
-            if (Value == 1)
-            {
-                list.Add(1076759, "{0}\t{1}.0 Skill Points", GetName(), Value);
-            }
-            else
-            {
-                list.Add(1076759, "{0}\t{1} Skill Points", GetName(), Value);
-            }
+            list.Add(1076759, $"{GetName()}\t{Value:0.#} Skill Points");
```
kamronbatman, line 39: This doesn't fix rounding, but the Value probably shouldn't be a decimal anyway.

PR #634, opened 2021-06-02
Added choice for extended status (only ML era)

romanlysenko:


kamronbatman (changes requested):
Let's create a variable and fill the value through an Initialize method. That way we don't have to read the config every time.

Let me know if you need help doing this.

PR #653, opened 2021-06-13
feat(core): Server configuration system

nullptr-w8:
Older video how to using https://vimeo.com/manage/videos/523271600

Weapon created using minIntensity to maxIntensity has 40% will be  regular it is original server logic, I deleted it from my server

Command to configure monsters is [mobmenu  
button ok is stats configure  (str, dex, int, dmg)   using original system roll after spawn minStr to maxStr.  
button arrow is loot system put gold and press arrow again  (at spawn time is create drop after spawn) 
count of gold random like  5d50+100 using original drop system. (using only count, not impossible, we can use only dice roll format)

Command to create loot pack is  [lp we can change drop in TreasureBoxChest 
create pack with name  TreasureBoxChest0 (is map 1 lvl)  put drop. 

Create Full Unique Monster  

step 1 create lootpack using  < [lp  > name like TheButcher
step  2  open stats editor using command < [lp s TheButcher >and configure it 
step 3 put to spawn properties  string like it :   hue 888 name "TheButcher" pack TheButcher


kamronbatman:
I think I might change this so that it can read multiple loot pack files instead of one giant loot pack json.

nullptr-w8:
> I think I might change this so that it can read multiple loot pack files instead of one giant loot pack json.

This is good idea by notepad edit ;) but I think the best way it's sqlite + web service for configurator server. 
This is way can do configure server for people technical less.

kamronbatman:
> > I think I might change this so that it can read multiple loot pack files instead of one giant loot pack json.
> 
> This is good idea by notepad edit ;) but I think the best way it's sqlite + web service for configurator server.
> This is way can do configure server for people technical less.

I agree, that is a big project. I definitely want full web based configuration system eventually.

Projects/UOContent/Gumps/Lootpack/PackWorker.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (fPack is not null)
+            {
+                try
+                {
+                    //copy list
+                    var Pack = fPack.LootItems?.ToList();
+                    if (Pack != null)
+                    {
+                        //include another pack to stack
+                        var DestPack = Pack.Where(_ => _.IsDestPack)
```
kamronbatman, line 57: We should eliminate LINQ. I'll take care of that.

Projects/UOContent/Gumps/Lootpack/PackWorker.cs
```diff
@@ -0,0 +1,83 @@
...
+                              .Select(_ => packLoader.GetPackLootByName(_.Name)).SelectMany(_ => _).ToList();
+                        Pack.AddRange(DestPack);
+                        //create a loot pack
+                        var Items = Pack.Where(_ => !_.IsDestPack).Select(_ => new LootPackEntry(_.AtSpawnTime,
+                       new LootPackItem[] { new(AssemblyHandler.FindTypeByName(_.TypeName), 1) },
+                       _.DropChance, _.Quantity
+                       , _.maxProps, _.minIntensity,
```
kamronbatman, line 64: This needs linting.

kamronbatman:
Let's continue the discussion at #1734 about how we want a new loot system to look. This is a pretty big subject!

PR #659, opened 2021-07-13
fix(build): Adds codeql-analysis.yml

kamronbatman:


kamronbatman:
It ran, and finished, and concluded that it timed out. Really weird.

PR #664, opened 2021-07-26
fix(core): Fixes a timer bug where Stop wasn't called internally

kamronbatman:


Projects/Server/Timer/Timer.TimerWheel.cs
```diff
@@ -107,23 +107,31 @@ private static void Execute(Timer timer)
...
-
+                // We remove it before `OnTick()` so time references can be nulled and returned to cache safely from within OnTick.
+                // This can be done by checking if Index < Count - 1 (still more iterations)
                 if (timer.Running)
                 {
                     RemoveTimer(timer);
 
-                    if (timer.Count == 0 || ++timer.Index < timer.Count)
+                    if (finished)
                     {
```
jaedan, line 119: For single-shot timers, finished will be false. But I think you still want to ensure timer.Running is set to false.

kamronbatman: Per conversation, one-shot timers should have a count of 1. Infinite intervals have a count of 0.

PR #676, opened 2021-08-07
Adds a PooledRefList

kamronbatman:
Pooled version of List<T>. Might not use it, we will see.


Projects/Server/Collections/PooledList.cs
```diff
@@ -0,0 +1,1245 @@
...
+                    {
+                        T[] newItems = ArrayPool<T>.Shared.Rent(_size);
+                        if (_size > 0)
+                        {
+                            Array.Copy(_items, newItems, _size);
+                        }
+
+                        if (_items.Length > 0)
+                        {
+                            ArrayPool<T>.Shared.Return(_items);
```
kamronbatman, line 120: Returns should clear the array so references are kept by the GC?

PR #692, opened 2021-08-17
fix(codegen): Codegens more armor items

kamronbatman:


kamronbatman:
This had a bad merge, so it was removed from `main`.

PR #698, opened 2021-08-18
fix(codegen): Codegens Weapons

mikedijkstra:
This PR codegens the remaining Weapons except for `Weapons/BaseWeapon.cs`. I didn't have time to finish that so it can be done in a follow up PR.


Projects/UOContent/Items/Weapons/BaseMeleeWeapon.cs
```diff
@@ -2,16 +2,13 @@
 
 namespace Server.Items
 {
-    public abstract class BaseMeleeWeapon : BaseWeapon
+    [Serializable(0, false)]
```
kamronbatman, line 5: I might need to come up with some kind of codegen for this scenario. 
This class actually doesn't serialize a version variable. It passes through entirely.

Projects/UOContent/Items/Weapons/PoleArms/BasePoleArm.cs
```diff
@@ -6,17 +6,20 @@
...
+        [SerializableFieldAttr("[CommandProperty(AccessLevel.GameMaster)]")]
+        private bool _showUsesRemaining;
 
-        private int m_UsesRemaining;
+        [SerializableField(1)]
+        [InvalidateProperties]
+        [SerializableFieldAttr("[CommandProperty(AccessLevel.GameMaster)]")]
+        private int _usesRemaining;
 
-        public BasePoleArm(int itemID) : base(itemID) => m_UsesRemaining = 150;
```
kamronbatman, line 15: This ctor was removed probably shouldn't have been.

Projects/UOContent/Items/Weapons/PoleArms/BasePoleArm.cs
```diff
@@ -6,17 +6,20 @@
 
 namespace Server.Items
 {
-    public abstract class BasePoleArm : BaseMeleeWeapon, IUsesRemaining
+    [Serializable(0)]
```
kamronbatman, line 9: This one is not encoded so we need a false.

Projects/UOContent/Items/Weapons/Swords/Scimitar.cs
```diff
@@ -1,15 +1,12 @@
 namespace Server.Items
 {
     [Flippable(0x13B6, 0x13B5)]
-    public class Scimitar : BaseSword
+    [Serializable(0)]
```
kamronbatman, line 4: This one should not be encoded.

Projects/UOContent/Items/Weapons/PoleArms/BasePoleArm.cs
mikedijkstra: I think this is still present, it's just the diff made it look like it was removed when it was updated. See line 22 below:
```
public BasePoleArm(int itemID) : base(itemID) => _usesRemaining = 150;
```

Projects/UOContent/Items/Weapons/BaseMeleeWeapon.cs
mikedijkstra: I've reverted this for now.

Issue #746, opened 2021-08-29
Quivers withouth damagebonus sets damage to 0


SpeedyDevil:
When wearing basequiver (elven quiver) then my dmg is 1 (because of fallback if(damage<0) damage = 1). When wearing any quiver that has set damage bonus, then dmg is right but damagetypes are bad.

In baseweapon there should be called GetDamageTypes() everytime and then "quiver.AlterBowDamage()" when player wears quiver (it adds bonuses to current damagetypes)

PR #747, opened 2021-08-29
Fixes damagetypes when wearing quivers

SpeedyDevil:
GetDamageTypes should be called everytime. Only when wearing quiver there should be alteredBowDamage into this types.


SpeedyDevil:
#746

kamronbatman (approved):

PR #776, opened 2021-09-08
fix: Uses client bug to make addon containers work

kamronbatman:
* Fixes tooltip for container furniture so it works. This doesn't work on ClassicUO.
* Fixes double click opening container furniture by double clicking the addon component.
* Fixes the context menu not showing up sometimes on the addon piece.

Screen_Shot_2021-09-12_at_4 27 28_PM

### DIFF
--- a/Projects/Server/Items/Container.cs
+++ b/Projects/Server/Items/Container.cs
@@ -763,7 +763,7 @@ namespace Server.Items
             {
                 if (Core.ML)
                 {
-                    if (ParentsContain<BankBox>()) // Root Parent is the Mobile.  Parent could be another containter.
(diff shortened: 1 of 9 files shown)


Arthrutus (commented):
Comment out the following in AddonContainerComponent: 

public override void GetProperties(ObjectPropertyList list)
        {
            // base.GetProperties(list);
            Addon?.GetProperties(list);
        }

This is not OSI/UODemise accurate. But keep it, just comment it out with a statement explaining what it does in case someone wants to add it in. Also keep base.GetProperties(list) commented out as well as it causes the highlighting issue.

PR #786, opened 2021-09-16
fix: Fixes A* Pathfinding

nullptr-w8:
Vertex don't  clear after a search, it's use old buffer.


Projects/UOContent/Engines/Pathing/FastAStarAlgorithm.cs
```diff
@@ -236,11 +232,11 @@ public override Direction[] Find(Mobile m, Map map, Point3D start, Point3D goal)
                     {
                         dirs[backtrack++] = path[--pathCount];
                     }
-
+
```
kamronbatman, line 235: Let's remove extra spaces.

PR #794, opened 2021-09-18
feat: Adds auto archiving

### LINKED ISSUE #791: Improve backup system
#### Start of every hour:
* Archive all saves for that hour
* Delete all saves except for the latest

#### Every day:
* Archive all saves (hourly) for that day
* Delete all saves except for the latest

#### Every 1st of the Month:
* Archive all saves for that month
* Delete all saves from that month

#### Configurations:
* Backup path. _defaults to `./Backups`_

#### Compression:
Archives should be compressed using [zstd](http://facebook.github.io/zstd/).

#### World Loading:
If the saves folder is empty or missing, look at Backup path for zst archives (named appropriately).
Then attempt to extract the latest archive and copy over the newest world save.
Should we prompt the user to do this instead of doing it automatically?

kamronbatman:
## Adds Auto Archiving
Backups are archived once an hour, day, and month. Archives older than 60 days are pruned automatically.

_Note: Automatic pruning is off by default_

### Archive compression format
The following formats are supported:
* Zstd (The fastest with best compression ratio)
* GZip
* Zip
* None (Tar)

_Note: By default archives use [zstandard](http://facebook.github.io/zstd/) format.
The archive format can be changed in modernuo.json `autoArchive.compressionFormat`_

### Restoring world from archive
Move the archive to the Saves folder (tar.zst file). On startup the server will extract the file and restore the latest save. See pictures below.

#### Windows
* Use the latest version of [7-zip w/ ZStandard](https://github.com/mcmilk/7-Zip-zstd/releases/latest)
  1. Extract the `.tar.zst` file.
  2. Extract the `.tar` file. (Yes you have to do it in two steps)
* On Windows 10 you can use the command line. `zstd.exe` is in the Assemblies folder after building ModernUO.
  1. `tar --use-compress-program "Distribution\Assemblies\zstd.exe -d" -xvf "Archives\Hourly\archivefile.tar.zst" -C "path to where you want to extract it"`
#### Mac
* Install [Keka](https://www.keka.io)
  1. Drop the .tar.zst onto the keka interface.
#### Linux
  1. Install zstd from a package manager
  2. Run `tar -I zstd -xvf "Archives\Hourly\archivefile.tar.zst" -C "path to where you want to extract it"`

### Other changes
* Changes Autosave to occur at the same time no matter when the server is booted.
* Adds `[SaveFrequency <delay> [warning]`command to set save frequency and warning frequency in-game.
* Adds support for time zones that are configurable. The system timezone can also be manually configured. Check `TimeZoneHandler.cs` for details.

Screen Shot 2021-09-22 at 11 23 18 PM
Screen Shot 2021-09-22 at 11 23 39 PM
Screen Shot 2021-09-22 at 11 34 39 PM
Screen Shot 2021-09-22 at 11 34 50 PM

Closes #791


Projects/UOContent/World Saves/AutoSave.cs
```diff
@@ -0,0 +1,132 @@
...
+                    BroadcastWarning();
+                    StartTimer(Warning, Save);
+                }
+                else
+                {
+                    Save();
+                }
+            }
+
+            _nextSave += Delay - Warning;
```
kamronbatman, line 93: This is wrong. It's already offset by `Warning`, so it only needs to be `_nextSave += Delay`

kamronbatman:
We should stop watch the restoration and archiving.

PR #795, opened 2021-09-18
Some minor fixes

SpeedyDevil:
Fixed Meditation skill check
Fixed eastToWest calc of magic fields

### DIFF
--- a/Projects/UOContent/Skills/Meditation.cs
+++ b/Projects/UOContent/Skills/Meditation.cs
@@ -71,7 +71,7 @@ namespace Server.SkillHandlers
             }
 
             var skillVal = m.Skills.Meditation.Value;
-            var chance = (50.0 + ((skillVal - (m.ManaMax - m.Mana)) * 2)) / 100;
+            var chance = (50.0 + (skillVal - (m.ManaMax - m.Mana)) * 2) / 100;
 
             if (chance > Utility.RandomDouble())
             {
--- a/Projects/UOContent/Skills/Meditation.cs
+++ b/Projects/UOContent/Skills/Meditation.cs
@@ -71,7 +71,7 @@ namespace Server.SkillHandlers
             }
 
             var skillVal = m.Skills.Meditation.Value;
-            var chance = (50.0 + (skillVal - (m.ManaMax - m.Mana)) * 2) / 100;
+            var chance = (50.0 + ((skillVal - (m.ManaMax - m.Mana)) * 2)) / 100;
 
             if (chance > Utility.RandomDouble())
             {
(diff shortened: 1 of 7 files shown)


Projects/UOContent/Skills/Meditation.cs
```diff
@@ -71,7 +71,7 @@ public static TimeSpan OnUse(Mobile m)
             }
 
             var skillVal = m.Skills.Meditation.Value;
-            var chance = (50.0 + (skillVal - (m.ManaMax - m.Mana)) * 2) / 100;
+            var chance = (50.0 + ((skillVal - (m.ManaMax - m.Mana)) * 2)) / 100;
```
kamronbatman, line 74: Double parenthesis, let's remove them.

Projects/UOContent/Spells/Base/SpellHelper.cs
```diff
@@ -220,7 +220,7 @@ public static bool CheckCombat(Mobile m)
                 }
             }
 
-            if (Core.Expansion == Expansion.AOS)
+            if (Core.AOS)
```
kamronbatman, line 223: I think this was supposed to specifically be for only AOS and not later?
Let's double check if possible what the different eras should be.

Projects/UOContent/Spells/Base/SpellHelper.cs
```diff
@@ -257,6 +257,17 @@ public static bool AdjustField(ref Point3D p, Map map, int height, bool mobsBloc
             return false;
         }
 
+
+        public static bool GetEastToWest(IPoint3D from,IPoint3D target)
```
kamronbatman, line 261: We should already have this somewhere I thought. I'll double check when I get back.

Projects/UOContent/Spells/Base/SpellHelper.cs
SpeedyDevil: ServUO doesnt check it for expansion at all.

Projects/UOContent/Spells/Base/SpellHelper.cs
SpeedyDevil: I didnt find it anywhere. In spells there are separately the condition (and not working :) )

Projects/UOContent/Skills/Meditation.cs
kamronbatman, line 74: I checked and I don't think this does anything.
Screen Shot 2021-09-18 at 8 28 11 PM

Note the gray parenthesis

SpeedyDevil: hmm ,I wonder why I edited it. ServUO have it with parenthesis but it doesnt change anything (only reading is better with it).

kamronbatman (approved):

PR #798, opened 2021-09-21
feat(expansions): Changes expansions to be data driven from expansion.json

michaelwhelehan:
* Adds a generic custom converter for boolean flags
* Adds an expansion.json config file for expansion features
* Loads expansion data from expansion.json instead of it being hardcoded


Projects/Server/ExpansionInfo.cs
```diff
@@ -141,105 +144,31 @@ public class ExpansionInfo
...
-            };
+                var expansion = expansions[i];
+                if (expansion.ClientVersion != null)
+                    Table[i] = new ExpansionInfo(
+                        i,
+                        expansion.Name,
+                        expansion.ClientVersion,
+                        expansion.FeatureFlags,
+                        expansion.CharacterListFlags,
+                        expansion.HousingFlags);
```
kamronbatman, line 162: Let's fix the alignment, the parenthesis should go on the next line.

Projects/Server/Json/Converters/ClientVersionConverter.cs
```diff
@@ -0,0 +1,37 @@
+/*************************************************************************
+ * ModernUO                                                              *
+ * Copyright 2019-2020 - ModernUO Development Team                       *
```
kamronbatman, line 3: Let's update this to 2021 (I admit, I need to update all of the files probably)

Projects/Server/Json/Converters/ClientVersionConverter.cs
```diff
@@ -0,0 +1,37 @@
...
+ * Email: hi@modernuo.com                                                *
+ * File: MapConverter.cs                                                 *
```
kamronbatman, line 5: Let's update this to ClientVersionConverter.cs.
I wonder if VS has an autocomplete macro like Rider does. 🤔

Projects/Server/Json/Converters/ClientVersionConverterFactory.cs
```diff
@@ -0,0 +1,29 @@
+/*************************************************************************
+ * ModernUO                                                              *
+ * Copyright 2019-2020 - ModernUO Development Team                       *
+ * Email: hi@modernuo.com                                                *
+ * File: Point2DConverterFactory.cs                                      *
```
kamronbatman, line 5: Same here

Projects/Server/Json/Converters/FlagsConverter.cs
```diff
@@ -0,0 +1,71 @@
+/*************************************************************************
+ * ModernUO                                                              *
+ * Copyright 2019-2020 - ModernUO Development Team                       *
+ * Email: hi@modernuo.com                                                *
+ * File: TimeSpanConverter.cs                                            *
```
kamronbatman, line 5: Same

Projects/Server/Json/Converters/FlagsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>. *
+ *************************************************************************/
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Server.Json
+{
+    public class FlagsConverter<T> : JsonConverter<T> where T : struct
```
kamronbatman, line 22: I think we can do `struct, Enum`.
I am also wondering if that helps us skip the (T)(object) double casting.

Projects/Server/Json/Converters/FlagsConverterFactory.cs
```diff
@@ -0,0 +1,29 @@
+/*************************************************************************
+ * ModernUO                                                              *
+ * Copyright 2019-2020 - ModernUO Development Team                       *
+ * Email: hi@modernuo.com                                                *
+ * File: TimeSpanConverterFactory.cs                                     *
```
kamronbatman, line 5: Same as above

Projects/Server/Json/Converters/FlagsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            return (T) (object) flags;
+        }
+
+        public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options)
+        {
+            writer.WriteStartObject();
+            foreach (var flagName in Enum.GetNames(typeof(T)))
+            {
+                if (!flagName.StartsWith("Expansion"))
```
kamronbatman, line 62: What is special about `Expansion` that we have to skip it?

Projects/Server/Json/Converters/FlagsConverter.cs
kamronbatman: Maybe something like this?

```cs
/*************************************************************************
 * ModernUO                                                              *
 * Copyright 2019-2021 - ModernUO Development Team                       *
 * Email: hi@modernuo.com                                                *
 * File: FlagsConverter.cs                                               *
 *                                                                       *
 * This program is free software: you can redistribute it and/or modify  *
 * it under the terms of the GNU General Public License as published by  *
 * the Free Software Foundation, either version 3 of the License, or     *
 * (at your option) any later version.                                   *
 *                                                                       *
 * You should have received a copy of the GNU General Public License     *
 * along with this program.  If not, see <http://www.gnu.org/licenses/>. *
 *************************************************************************/

using System;
using System.Runtime.CompilerServices;
using System.Text.Json;
using System.Text.Json.Serialization;

namespace Server.Json
{
    public class FlagsConverter<T> : JsonConverter<T> where T : struct, Enum
    {
        public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            var flags = 0ul;
            var underlyingType = Enum.GetUnderlyingType(typeof(T));

            while (true)
            {
                reader.Read();
                if (reader.TokenType == JsonTokenType.EndObject)
                {
                    break;
                }

                if (reader.TokenType != JsonTokenType.PropertyName)
                {
                    throw new JsonException("Invalid Json structure for Flag object");
                }

                var key = reader.GetString();

                reader.Read();

                if (!reader.GetBoolean() || !Enum.TryParse<T>(key, out var val))
                {
                    continue;
                }

                flags |= ConvertToUInt64(underlyingType, val);
            }

            switch (Type.GetTypeCode(underlyingType))
            {
                case TypeCode.SByte:
                    {
                        sbyte num = (sbyte)flags;
                        return Unsafe.As<sbyte, T>(ref num);
                    }
                case TypeCode.Byte:
                    {
                        byte num = (byte)flags;
                        return Unsafe.As<byte, T>(ref num);
                    }
                case TypeCode.Int16:
                    {
                        short num = (short)flags;
                        return Unsafe.As<short, T>(ref num);
                    }
                case TypeCode.UInt16:
                    {
                        ushort num = (ushort)flags;
                        return Unsafe.As<ushort, T>(ref num);
                    }
                default:
                case TypeCode.Int32:
                    {
                        int num = (int)flags;
                        return Unsafe.As<int, T>(ref num);
                    }
                case TypeCode.UInt32:
                    {
                        uint num = (uint)flags;
                        return Unsafe.As<uint, T>(ref num);
                    }
                case TypeCode.Int64:
                    {
                        long num = (long)flags;
                        return Unsafe.As<long, T>(ref num);
                    }
                case TypeCode.UInt64:
                    {
                        return Unsafe.As<ulong, T>(ref flags);
                    }
            }
        }

        private static ulong ConvertToUInt64(Type underlyingType, object value) =>
            Type.GetTypeCode(underlyingType) switch
            {
                TypeCode.SByte  => (ulong)(sbyte)value,
                TypeCode.Byte   => (byte)value,
                TypeCode.Int16  => (ulong)(short)value,
                TypeCode.UInt16 => (ushort)value,
                TypeCode.Int32  => (ulong)(int)value,
                TypeCode.UInt32 => (uint)value,
                TypeCode.Int64  => (ulong)(long)value,
                TypeCode.UInt64 => (ulong)value,
                _               => throw new InvalidOperationException()
            };

        public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options)
        {
            writer.WriteStartObject();
            var underlyingType = Enum.GetUnderlyingType(typeof(T));
            var intValue = ConvertToUInt64(underlyingType, value);

            foreach (var flagName in Enum.GetNames(typeof(T)))
            {
                var flag = ConvertToUInt64(underlyingType, Enum.Parse<T>(flagName, false));
                writer.WriteBoolean(flagName, (intValue & flag) != 0);
            }
            writer.WriteEndObject();
        }
    }
}
```

kamronbatman (changes requested):

Projects/Server/Json/Converters/ClientVersionConverter.cs
michaelwhelehan: I use Rider haha can you show me how it works?

Projects/Server/Json/Converters/FlagsConverter.cs
michaelwhelehan: The expansion flags are just aggregates of the individual flags so I don't want to write those flags out to booleans

Projects/Server/ExpansionInfo.cs
michaelwhelehan: Done

Projects/Server/Json/Converters/FlagsConverter.cs
michaelwhelehan: Ahh thanks for providing this, I've updated it and tested that it works. Although it feels a lot more verbose than the double cast it is probably a cleaner solution in the end

Projects/Server/Json/Converters/ClientVersionConverter.cs
kamronbatman: @michaelwhelehan import the https://github.com/modernuo/ModernUO/blob/main/rider-settings.zip file into your settings.

Then you type `copy` and will see a dropdown for copyright, press enter, and it is done for you.

Projects/Server/Json/Converters/FlagsConverter.cs
kamronbatman: So let's do `Enum.GetValues(typeof(T))` to get the values in an array.
Then let's check it generically:
```cs
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        private static int GetUnderlyingTypeLength(TypeCode typeCode) =>
            typeCode switch
            {
                TypeCode.Byte   => 1,
                TypeCode.SByte  => 1,
                TypeCode.Int16  => 2,
                TypeCode.UInt16 => 2,
                TypeCode.Char   => 2,
                TypeCode.Int32  => 4,
                TypeCode.UInt32 => 4,
                TypeCode.Int64  => 8,
                TypeCode.UInt64 => 8,
                _               => 8
            };
```

```cs
var size = GetUnderlyingTypeLength(underlyingType) - 1;
...
var flag = ConvertToUInt64(underlyingType, flagValues[i]);
if (flag == 0 || (flag & size) == 0)
{
  // write value
}
```

Projects/Server/ExpansionInfo.cs
```diff
@@ -331,4 +263,23 @@ public static ExpansionInfo GetInfo(int ex)
 
         public override string ToString() => Name;
     }
+
+    public record ExpansionConfig
+    {
+        public string Name { get; init; }
+
+        [JsonConverter(typeof(ClientVersionConverter))]
```
kamronbatman, line 271: Client version can go into options. Only enum flags should be a custom attributed converter.

Projects/Server/ExpansionInfo.cs
```diff
@@ -141,105 +145,33 @@ public class ExpansionInfo
...
-                    11,
-                    "Endless Journey",
-                    new ClientVersion("7.0.61.0"),
-                    FeatureFlags.ExpansionEJ,
-                    CharacterListFlags.ExpansionEJ,
-                    HousingFlags.HousingEJ
-                )
-            };
+                var expansion = expansions[i];
+                if (expansion.ClientVersion != null)
```
kamronbatman, line 156: Almost missed this. We should always have braces around all if statements, even single line.

kamronbatman:
I am not sure why but the Table is empty during unit tests? We should look into this before merging.

kamronbatman (approved):

PR #807, opened 2021-09-27
[PoC] Add WebRender

nullptr-w8:
Webrender


kamronbatman:
This is great!
I think this will become impractical without performance issues. Eventually we will want to provide html from CDN.
Maybe we can simply provide a URL?

nullptr-w8:
> This is great!
> I think this will become impractical without performance issues. Eventually we will want to provide html from CDN.
> Maybe we can simply provide a URL?

Yeah, we can use URL, but for now only content we use original socket.  At the moment, I try to swap sciter(only basic JS it's use for example gump)  sctier.JS(vue,react,etc) sctier

kamronbatman:
@nullptr-w8 Let's move forward with this in some fashion:
@michaelwhelehan and @jabinb, I would love to know your thoughts. I see this working out by doing the following:

1. Build gumps using HTML/CSS/JS.
2. Generate the final bundle of files.
3. Host them archived, or individually on CDN.
4. In-game gumps will be a URL to that bundle, which would then be cached and versioned.
5. Gumps send additional dynamic data (strings and potentially the rest) via the rest of the gump packet.

paezao:
@kamronbatman should we close this at this point?

kamronbatman:
I was keeping this open as a PoC. Let's put it into draft.

PR #821, opened 2021-10-08
Epoll: set struct pack and size

jabinb:
This seems to fix the memory corruption in `EPollgroup.cs` when there's more than one event in the queue. Tested with 3 clients on Linux x64 5.13.0, libc 2.33. 

I can't test on Windows so you please check before merging.

### Before
image

### After
image


kamronbatman:
This will work on Windows, just needs the recompiled version. I'll test today or tomorrow.

Thank you!

PR #834, opened 2021-10-23
fix: Removes LINQ from Map.cs

kamronbatman:


Projects/Server/Maps/Map.cs
```diff
@@ -555,9 +624,10 @@ public void FixColumn(int x, int y)
 
             GetAverageZ(x, y, out _, out var landAvg, out _);
 
-            var items = AcquireFixItems(this, x, y);
+            var items = ArrayPool<Item>.Shared.Rent(128);
```
kamronbatman, line 627: Replace with single threaded pool of Arrays of 128 items?
This should be an Array of 16 Arrays.
Then log a warning if we need to increase the pool.
To that end, set up a DEBUG mode to identify the code paths with nested pool checks.

PR #842, opened 2021-11-11
fix: Fixes animal taming timeout in case target is cancelled.

nullptr-w8:


kamronbatman:
This is timeout in case the target is cancelled?
I think 30 seconds makes sense. Let's Remove the comment.

Thank!

PR #845, opened 2021-11-13
House items decay fix

nullptr-w8:


kamronbatman:
@nullptr-w8 What is this fixing?
I think the logic is supposed to say:
Item can decay if it is not inside the building, or it is not locked down and it is not secured.
```cs
        public override bool OnDecay( Item item ) =>
            !House.IsInside(item) || !House.HasLockedDownItem(item) && !House.HasSecureItem(item) && base.OnDecay(item);
```

This should work. If you are having issues, please let me know so I can investigate.

kamronbatman:
Per our conversation, this is not a bug but instead a preference.
On OSI items decay unless they are locked down or secured.

I think it's an ok idea to leave a comment or make a configuration but I don't want to change the default behavior right now.

PR #865, opened 2021-11-27
Guilds fix

nullptr-w8:
Item's read entity, <guilds> but it didn't Deserialize


SpeedyDevil (approved):
Works like a charm :)

kamronbatman:
Even though this may work, it technically is not the correct fix. I'll look into this more.

PR #866, opened 2021-11-28
fix - Spells CheckXSequence crashes

SpeedyDevil:
After spell cleanup, CheckHSequence and CheckBSequence crashes when target is null (eg. died after click on target but before spell cast).


kamronbatman:
Thanks! I think this was covered up by the duplicate null checks that I ended up removing.

PR #882, opened 2021-12-07
fix: Fixes VS builds by reverting back to nstd20

kamronbatman:
* Adds explicit dependencies VS requires
* Reverts back to .NET Standard 2.0


3HMonkey (approved):
Tested and works like charm

PR #890, opened 2021-12-20
fix - mageryspell circle out of bounds

SpeedyDevil:
When spell from first circle is casted from scroll, it crashed. Spell circle cant be below 0.

### DIFF
--- a/Projects/UOContent/Spells/Base/MagerySpell.cs
+++ b/Projects/UOContent/Spells/Base/MagerySpell.cs
@@ -30,6 +30,8 @@ namespace Server.Spells
                 circle -= 2;
             }
 
+            circle = Math.Max(0, circle);
+
             // Original RunUO algorithm for required skill
             // const double chanceOffset = 20.0
             // const double chanceLength = 100.0 / 7.0


Projects/UOContent/Spells/Base/MagerySpell.cs
```diff
@@ -30,6 +30,8 @@ public override void GetCastSkills(out double min, out double max)
                 circle -= 2;
             }
 
+            circle = Math.Max(0, circle);
```
kamronbatman, line 33: This actually makes first/second circles too easy. We want them to essentially never fail. I wonder if min/max of 0 is good enough? We should test this:

The math is `-7` for each negative circle for ML+, and `-5` for each negative circle that is earlier than ML.

I'll think about a good way to do it. There are multiple options here.

kamronbatman: @SpeedyDevil
```cs
        private static readonly double[] _requiredSkill = Core.ML ?
            new[] { -28.0, -14.0, 0.0, -4.0, 10.0, 24.0, 38.0, 52.0, 66.0, 80.0 } :
            new[] { -20.0, -10.0, 0.0, 10.0, 20.0, 30.0, 40.0, 50.0, 60.0, 70.0 };


        public override void GetCastSkills(out double min, out double max)
        {
            // Index starts at circle -2 (to account for scrolls)
            var circle = (int)Circle;

            if (Scroll == null)
            {
                circle += 2;
            }

            // Original RunUO algorithm for required skill
            // const double chanceOffset = 20.0
            // const double chanceLength = 100.0 / 7.0
            // var avg = chanceLength * circle;
            // min = avg - chanceOffset;
            // max = avg + chanceOffset;

            // Correct algorithm according to OSI.
            // TODO: Verify this algorithm on OSI for latest expansion.
            min = _requiredSkill[circle];
            max = min + 40;
        }
```

Thoughts on this?

PR #920, opened 2022-01-30
Fix globalization invariant mode 

stefanomerotta:
In some cases the Globalization Invariant mode is enabled by default even if not specified.

Starting from .NET 6 there are some breaking changes that causes exceptions when this mode is active and the code try to get a specific CultureInfo.

I made the configuration explicit for server project and remove two hard coded culture info found in the UOContent project.

PR #921, opened 2022-01-30
fix: Fixes en-us forced pricing culture throwing when bulding in VS

kamronbatman:
Fixes culture not found exception with global invariance set through builds in VS.

Closes #920 as a more proper fix.


Arutosio:
i did the build with visual studio and launched the server from visual studio 2022.
now i am no longer disconnected when i open my seller's backpack

Arutosio (approved):

PR #925, opened 2022-02-04
Fix missing EJ houses in catalog

### LINKED ISSUE #923: [Bug] House - House Limit
House: "The Robin's Roost Castle"
image
image
i placed a house with the gm then i transferred the property to a player then i added some wall and two doors (with property moviable false) with the gm. 
I don't know if the command worked on this house before, I haven't tried.
after the player was unable to secure a trunk with the command

System msg: this action would exceed the lockdown limit of this house

stefanomerotta:
Fix missing EJ houses in AOS catalog: this causes max locked down and securable to zero and prevent users to lock or secure items and containers in owned houses.

resolve #923

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@
 .DS_Store
 
 /packages/*
+/Distribution/Configuration/server-access.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -32,3 +32,4 @@
 .DS_Store
 
 /packages/*
+/Distribution/Configuration/server-access.json
```
kamronbatman, line 35: Good call on this

kamronbatman (approved):

PR #931, opened 2022-02-13
ML Peerless Additions

Arthrutus:
PR focuses on the following:

1. Teleporter location corrections
2. Addition of ML bosses
3. Incorporation of ItemSocket for tracking timers.

Very rough PR and will need heavy editing in order to function properly.

### DIFF
--- a/Distribution/Data/teleporters.json
+++ b/Distribution/Data/teleporters.json
@@ -40,9 +40,9 @@
     "back": true
   },
   {
-    "src": { "map": "Felucca", "loc": [589, 1637, 0] },
-    "dst": { "map": "Felucca", "loc": [6575, 888, 0] },
-    "back": false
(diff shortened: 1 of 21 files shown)


kamronbatman:
Let's use this as a guide, but we do not want to just outright copy from ServUO.

Distribution/Data/teleporters.json
```diff
@@ -40,9 +40,9 @@
     "back": true
   },
   {
-    "src": { "map": "Felucca", "loc": [589, 1637, 0] },
-    "dst": { "map": "Felucca", "loc": [6575, 888, 0] },
-    "back": false
+    "src": { "map": "Felucca", "loc": [588, 1637, 0] },
```
kamronbatman, line 43: Let's make a separate PR for the teleporter updates so we can merge those quickly.

PR #940, opened 2022-02-21
fix: House lock downs and secure ownership bugs

Arthrutus:
Currently reworking house lockdowns/Secures. This closely resemble OSI but still needs fixes before release.

1. When a co-owner or owner locks down an item, access levels below them should not be able to release them (E.g., an owner locks down an item, the co-owner should not be able to release said item).
2. When a player secures a container, the security access gump should list that player as an owner of the container. As it stands, the code currently defaults the house owner as the owner. This causes co-owners who select the "Owner Only" option to lose access to the container and become unable to release it.


Projects/UOContent/Multis/Houses/BaseHouse.cs
```diff
@@ -2070,7 +2070,7 @@ private void SetLockdown(Item i, bool locked, bool checkContains = false)
 
         public bool LockDown(Mobile m, Item item, bool checkIsInside)
         {
-            if (!IsCoOwner(m) || !IsActive)
+            if (!IsFriend(m) || !IsActive)
```
kamronbatman, line 2073: I am not sure if this is right. Let's do some testing.

Projects/UOContent/Multis/Houses/BaseHouse.cs
```diff
@@ -2408,7 +2423,17 @@ public void AddSecure(Mobile m, Item item)
                 }
                 else
                 {
-                    info = new SecureInfo((Container)item, SecureLevel.Owner);
+                    /* There seems to be some confusion regarding security access on secured containers.
```
kamronbatman, line 2426: The owner of the house should have access to everything.
This is both necessary and a problem since owners can hijack co-owner/friend property.
I am not sure this can be resolved.

Arthrutus: I agree. Owners will automatically have access to everyone. Perhaps we can make a new securelevel.HouseOwner?

Projects/UOContent/Multis/Houses/BaseHouse.cs
Arthrutus: Friends of the house have always been allowed to lock things down. This was tested via OSI and is at the link: https://uo.com/wiki/ultima-online-wiki/gameplay/houses-placing-a-house/house-ownership-4-managing-your-home/

Projects/UOContent/Multis/Houses/BaseHouse.cs
Arthrutus: Maybe we need to separate the security access levels and the house access levels. These are obviously not interchangeable. The security access levels only apply to each individual container in question and are applicable to the individual who secured said container, NOT the house owner (although they have access/control over it regardless).

For things like doors and teleporters, the owner is automatically the house owner.

Projects/UOContent/Multis/Houses/BaseHouse.cs
```diff
@@ -2596,7 +2625,7 @@ public void Kick(Mobile from, Mobile targ)
                 /* You have been ejected from this house.
                  * If you persist in entering, you may be banned from the house.
                  */
-                targ.SendLocalizedMessage(501341);
+                    targ.SendLocalizedMessage(501341);
```
kamronbatman, line 2628:
```suggestion
                targ.SendLocalizedMessage(501341);
```

kamronbatman:
It feels like the first item on the list is a bug. I agree that this should be fixed if that is the case. The _default_ security level should be the highest for that person's current access. If they are a friend, it is friend. If they are a co-owner, it is co-owner, etc.

The second one is also clearly a bug.

Let's move forward with this. I'll find some time to look into this more and clean up this PR.

PR #946, opened 2022-02-28
fix: Codegens containers

kamronbatman:


Projects/UOContent/Items/Containers/Container.cs
```diff
@@ -4,1013 +4,505 @@
 using Server.Multis;
 using Server.Network;
 
-namespace Server.Items
+namespace Server.Items;
+
+[ManualDirtyChecking]
```
kamronbatman, line 9: This is necessary since we can't codegen the class. Eventually MUO will throw an error without this. It doesn't do anything other than to tell the dirty tracker (when in strict mode) to ignore this class.

PR #953, opened 2022-03-04
fix: Fixes packet length checks

kamronbatman:
Fixes an issue with DropReq where an old client was sending in 14 bytes, but the server was expecting 15 bytes.

To fix this we introduced a new packet handler, `ContainerGridPacketHandler` and changed the code to determine the length of the packet dynamically using `GetLength(NetState)`.

Also fixed throttling so dropped packets are properly skipped.

### DIFF
--- a/Projects/Server/Network/NetState/NetState.cs
+++ b/Projects/Server/Network/NetState/NetState.cs
@@ -223,32 +223,20 @@ public partial class NetState : IComparable<NetState>
 
     private void SetPacketTime(int packetID)
     {
-        if (packetID is < 0 or >= 0x100)
+        if (packetID is >= 0 and < 0x100)
         {
-            return;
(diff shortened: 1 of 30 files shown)


Projects/Server/Network/NetState/NetState.cs
```diff
@@ -473,630 +473,629 @@ public bool GetSendBuffer(out CircularBuffer<byte> cBuffer)
...
+                            {
+                                _parserState = ParserState.ProcessingPacket;
+                                _parserState = HandlePacket(packetReader, packetId, out packetLength);
+                                break;
+                            }
                     }
                 }
 
-                reader.Commit();
+                if (_parserState is ParserState.AwaitingNextPacket or ParserState.Throttled)
```
kamronbatman, line 727: This should fix throttling.

PR #962, opened 2022-03-17
Dynamic region

nullptr-w8:
Create a dynamic custom region from game

[add regioncontroller

Region bound  (height, width) can be only  positive number, if we pass negative region doesn't work.
Format for ExcludeSpell  Wall of Stone,Clumsy,etc like name from the spellbook


kamronbatman:
We can do this without an item. 🤔

Projects/UOContent/Gumps/Props/PropsGump.cs
```diff
@@ -55,7 +55,8 @@ public class PropertiesGump : Gump
         protected readonly Type m_Type;
         protected int m_Page;
         protected int m_EntryCount;
-
+        public delegate void EditHandler(object m_Object);
```
kamronbatman, line 58: Can use `Action<object>` now.

Projects/Server/Data/Store/IRegistredStore.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                return result;
+            }
+
+            return default;
+        }
+
+        public virtual void Save()
+        {
+            JsonConfig.Serialize(_path, _data);
```
kamronbatman, line 80: I think we should queue changes to be written to disk asynchronously by a separate thread.

Projects/Server/Data/Loader.cs
```diff
@@ -0,0 +1,42 @@
...
+            Utility.PushColor(ConsoleColor.Green);
+            Console.Write("done");
+            Utility.PopColor();
+            Console.WriteLine("({0:F2} seconds)", stopwatch.Elapsed.TotalSeconds);
+        }
+       
+
+        private static void ConfigureServices(IServiceCollection services)
+        {
+            services.AddSingleton<IRegionStore>(new RegionStore("Data/CustomRegion.json"));
```
kamronbatman, line 39: I like this. In the future we should make this dynamic. Maybe 1 file per map?

Projects/Server/Data/Model/RegionModel.cs
```diff
@@ -0,0 +1,86 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.Json.Serialization;
+using System.Threading.Tasks;
+
+namespace Server
+{
+    public static partial class DataModel
```
kamronbatman, line 10: Requires `[Parsable]`?

Projects/UOContent/Gumps/ObjectViewer/BaseViewGump.cs
```diff
@@ -0,0 +1,170 @@
...
+        public override void OnResponse(NetState state, RelayInfo info)
+        {
+            var buttonID = info.ButtonID - 1;
+            var type = buttonID % _typeCount;
+            var index = buttonID / _typeCount;
+
+            switch (type)
+            {
+                case 0: // close
+                case -1: // close
```
kamronbatman, line 131: Consolidate `case 0` and `case -1` into `default` so that all invalid buttons will close the gump?

Projects/UOContent/Gumps/ObjectViewer/RegionView.cs
```diff
@@ -0,0 +1,124 @@
...
+                return;
+            }
+
+            RegionStore.Add(name, new RegionModel() { Name = name });
+            ReOpen();
+        }
+
+        public override void OnPressTwoButton(Mobile m, int index)
+        {
+            if (Collection.ElementAt(index).Value is RegionModel region)
```
kamronbatman, line 43: Is this always true?, I assume `Collection` is `List<KeyValuePair<string, RegionModel>>`.
```cs
var region = Collection[index].Value;
```

Projects/UOContent/Gumps/ObjectViewer/BaseViewGump.cs
nullptr-w8: var buttonID = info.ButtonID - 1;  // -1 == 0 
 var type = buttonID % _typeCount;

Projects/Server/Data/Loader.cs
nullptr-w8: I think for future, we can use db  we can store vendor items, loot, stats etc. 
I try to include sqlite code first with ef migration.

kamronbatman: The database is ok for data loaded during boot up and data that does not change often.

mdodkins:
oooo, I was going to steal the CustomRegion from Angel Island but I think you might be doing something similar here! cool!

mdodkins:
oh, it hasn't been updated for ages, that is sad

kamronbatman:
@mdodkins - I actually have an up-to-date and working regions controller for MUO. I'll add it here when I get time:
https://github.com/modernuo/Scripts/tree/main/Scripts

PR #971, opened 2022-03-26
fix: Adds Gen2 callback for each STArrayPool

kamronbatman:
Adds. Gen2Callback for STArrayPool so it can purge the internal array stacks properly. This only happens if significant timed has passed, 2 Gen 2's have been run, and memory pressure exceeds a threshold.

### DIFF
--- a/Projects/Server/Buffers/STArrayPool.cs
+++ b/Projects/Server/Buffers/STArrayPool.cs
@@ -6,6 +6,7 @@ using System.Buffers;
 using System.Diagnostics;
 using System.Numerics;
 using System.Runtime.CompilerServices;
+using System.Threading;
 
 namespace Server.Buffers;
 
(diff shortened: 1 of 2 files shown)


Projects/Server/Buffers/STArrayPool.cs
```diff
@@ -172,6 +174,12 @@ private STArray[] InitializeBuckets()
     {
         Debug.Assert(_cacheBuckets is null, $"Non-null {nameof(_cacheBuckets)}");
         var buckets = new STArray[BucketCount];
+
+        if (Interlocked.Exchange(ref _trimCallbackCreated, 1) == 0)
+        {
+            Gen2GcCallback.Register(o => ((STArrayPool<T>)o).Trim(), this);
```
kamronbatman, line 180: We could probably optimize this later by unregistering it if the stacks are empty. It isn't a big concern though since it is a single object per generic type.

PR #986, opened 2022-04-03
fix: Properly count timers for DumpInfo

jkachhad:
Resolves an issue where only the root timer was counted in the ring slot.


kamronbatman (approved):

kamronbatman:
Thank you!

PR #987, opened 2022-04-04
fix: Fixes food stacking for poisoned food.

Mink80:


kamronbatman:
The zero arg constructors aren't needed since the current ones have an optional argument.

kamronbatman:
@Mink80 per our discussion, creating objects should use the ModernUO extension method. Example:
```cs
var dup = obj.GetType().CreateInstance<Type>();
```

This will find a constructor, even if it has optional parameters, and use it. The reason why Activator.CreateInstance doesn't work is because optional parameters have a `Type.Missing`.

So you would need to do something like this:

```cs
(T)Activator.CreateInstance(
    typeof(T), 
    BindingFlags.CreateInstance |
    BindingFlags.Public |
    BindingFlags.Instance | 
    BindingFlags.OptionalParamBinding,
    null,
    new object[] {Type.Missing },
    CultureInfo.CurrentCulture
);
```

per this [stackoverflow post](https://stackoverflow.com/questions/11002523/activator-createinstance-with-optional-parameters)

I don't recommend doing this since allocating the object array, using the old .NET 2 activator, etc is really slow and obtuse.

kamronbatman (approved):

PR #1008, opened 2022-04-26
Travel Restrictions xml settings

DragonSlayer62:
This adds the abilities to be able to edit the restrictions of travel on the fly without having to get into the spell helper.

The old system in spellhelper.cs will need to be removed so there is no conflicts.


Projects/UOContent/Spells/Base/SpellHelper.cs
```diff
@@ -80,6 +83,116 @@ public static class SpellHelper
             1, 1
         };
 
+        public static void Configure()
+        {// Custom Travel Settings. Needs converted to Json in time.
+            Console.Write("SpellHelper: Loading TravelRestrictions...");
```
kamronbatman, line 88: Let's change this to use logger.

kamronbatman:
Let's focus on eliminating travel restrictions entirely and put the properties directly on the regions.

kamronbatman:
Let's discuss this on #1733 and revisit this with a better implementation.

PR #1010, opened 2022-05-02
fix: nullptr exception

tellundro:
fix for exception on line 364 due to _lifted hashmap not being initialized.

### DIFF
--- a/Projects/UOContent/Items/Containers/TreasureMapChest.cs
+++ b/Projects/UOContent/Items/Containers/TreasureMapChest.cs
@@ -44,7 +44,7 @@ public partial class TreasureMapChest : LockableContainer
     [Tidy]
     [SerializableField(5, setter: "private")]
     [SerializableFieldAttr("[CommandProperty(AccessLevel.GameMaster)]")]
-    private HashSet<Item> _lifted = new HashSet<Item>();
+    private HashSet<Item> _lifted;
 
     [Constructible]
     public TreasureMapChest(int level) : this(null, level)
@@ -361,6 +361,11 @@ public partial class TreasureMapChest : LockableContainer
 
     public override void OnItemLifted(Mobile from, Item item)
     {
+        if (_lifted == null)
+        {
+            _lifted = new HashSet<Item>();
+        }
+
         var notYetLifted = !_lifted.Contains(item);
 
         from.RevealingAction();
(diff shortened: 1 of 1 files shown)


kamronbatman:
@tellundro, I am thinking we should do a null check instead, and then initialize before adding. This way we aren't allocating for the HashSet until we need it.

tellundro:
@kamronbatman yeah that makes sense. I'll edit the code and push it again.

Projects/UOContent/Items/Containers/TreasureMapChest.cs
```diff
@@ -361,6 +361,11 @@ private bool CheckLoot(Mobile m, bool criminalAction)
 
     public override void OnItemLifted(Mobile from, Item item)
     {
+        if (_lifted == null)
```
kamronbatman, line 364: Let's change to this to:
```cs
_lifted ??= new HashSet<Item>();
```

And move it right above this line:
```cs
_lifted.Add(item);
```

kamronbatman (changes requested):

Projects/UOContent/Items/Containers/TreasureMapChest.cs
Tultra: @kamronbatman I guess we need to keep it above 

`var notYetLifted = !_lifted.Contains(item);`

which is where the exception happened.

Projects/UOContent/Items/Containers/TreasureMapChest.cs
```diff
@@ -361,7 +361,10 @@ private bool CheckLoot(Mobile m, bool criminalAction)
 
     public override void OnItemLifted(Mobile from, Item item)
     {
-        var notYetLifted = !_lifted.Contains(item);
+        
+        _lifted ??= new HashSet<Item>();
+        
+        var notYetLifted = !_lifted.Contains(item);
```
kamronbatman, line 367: Let's add a null check here, and move the assignment to where we need it (above the add):

```cs
var notYetLifted = _lifted?.Contains(item) != true;
```

Projects/UOContent/Items/Containers/TreasureMapChest.cs
```diff
@@ -361,15 +361,12 @@ private bool CheckLoot(Mobile m, bool criminalAction)
...
-        _lifted ??= new HashSet<Item>();
-        
-        var notYetLifted = !_lifted.Contains(item); 
-
+        var notYetLifted = _lifted?.Contains(item) != true; 
         from.RevealingAction();
 
         if (notYetLifted)
         {
+            _lifted = new HashSet<Item>();
```
kamronbatman, line 369: This needs to be `??=`

kamronbatman (approved):

PR #1033, opened 2022-05-27
fix: Adds multithread optional flag to ValueStringBuilder

### LINKED ISSUE #1032: Verify serialization methods exception
```cs
private static void VerifySerialization(Assembly assembly)
        {
            if (assembly != null)
            {
                Parallel.ForEach(assembly.GetTypes(), VerifyType);
            }
        }
```

The serialization verification is multithread but inside the **VerifyType** method a ValueStringBuilder is instantiated: it uses the single thread array pool that could causes exceptions if multiple assemblies are checked.

```cs
public ValueStringBuilder(int initialCapacity)
    {
        _arrayToReturnToPool = STArrayPool<char>.Shared.Rent(initialCapacity);
        _chars = _arrayToReturnToPool;
        _length = 0;
    }
```

We could use the standard StringBuilder in this scenario, as it is executed only in startup phase and additional memory allocation is not a problem, or add to ValueStringBuilder an option to select which pool it should use.

stefanomerotta:
Added multithread optional flag to ValueStringBuilder for multithread contexts.
Enabled new flag for VerifyType method.

Resolves #1032

### DIFF
--- a/Projects/Server/Buffers/ValueStringBuilder.cs
+++ b/Projects/Server/Buffers/ValueStringBuilder.cs
@@ -3,6 +3,7 @@
 
 #nullable enable
 using System;
+using System.Buffers;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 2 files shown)


Projects/Server/Buffers/ValueStringBuilder.cs
```diff
@@ -13,35 +14,45 @@ namespace Server.Buffers;
     private char[] _arrayToReturnToPool;
     private Span<char> _chars;
     private int _length;
+    private bool _mt;
 
-    public ValueStringBuilder() : this(64)
+    private ArrayPool<char> ArrayPool
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get => _mt ? ArrayPool<char>.Shared : STArrayPool<char>.Shared;
```
kamronbatman, line 22: This is great, we should refactor this (in another PR) for RefQueue and RefList

kamronbatman (approved):
This is awesome, thank you!

PR #1036, opened 2022-05-28
fix: Adds basic KR/EC support

kamronbatman:
- [X] Adds basic KR/EC login support
- [X] Adds support for EC client versions
- [X] Fixes a few minor bugs

Part of this was handled in #1506 , #1507


kamronbatman:
Let's revisit this another time.

PR #1063, opened 2022-06-14
Replaced throttlers and packet callbacks with function pointers

stefanomerotta:
Attempt to lower memory consumption and gain packet handling speed by replacing traditional delegates with function pointers.

For packet handlers callbacks that are static methods under the hood we don't need multicast support and we can remove the additional memory allocation for the standard delegate.
Function pointers should be faster than traditional delegates due to the different IL instruction and no invocation list.

### DIFF
--- a/Projects/Server/Network/ContainerGridPacketHandler.cs
+++ b/Projects/Server/Network/ContainerGridPacketHandler.cs
@@ -15,9 +15,10 @@
 
 namespace Server.Network;
 
-public class ContainerGridPacketHandler : PacketHandler
+public unsafe class ContainerGridPacketHandler : PacketHandler
 {
-    public ContainerGridPacketHandler(int packetID, int length, bool ingame, OnPacketReceive onReceive)
+    public ContainerGridPacketHandler(int packetID, int length, bool ingame,
+        delegate* managed<NetState, CircularBufferReader, int, void> onReceive)
         : base(packetID, length, ingame, onReceive)
     {
     }
--- a/Projects/Server/Network/NetState/NetState.cs
+++ b/Projects/Server/Network/NetState/NetState.cs
@@ -745,7 +745,7 @@ public partial class NetState : IComparable<NetState>
      * length is the total buffer length. We might be able to use packetReader.Capacity() instead.
      * packetLength is the length of the packet that this function actually found.
      */
-    private ParserState HandlePacket(CircularBufferReader packetReader, byte packetId, out int packetLength)
+    private unsafe ParserState HandlePacket(CircularBufferReader packetReader, byte packetId, out int packetLength)
     {
         PacketHandler handler = IncomingPackets.GetHandler(packetId);
         int length = packetReader.Length;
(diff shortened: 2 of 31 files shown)


stefanomerotta:
The only not-so-clear part is the ProtocolExtensions handler registration: I used a static array of handlers inside a static generic class that allowed me to maintain backward compatibility while using static method as callback

kamronbatman:
This is an interesting concept!!! I would love to see a benchmark. 🤔

stefanomerotta:
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
11th Gen Intel Core i5-1145G7 2.60GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.300
  [Host]     : .NET 6.0.5 (6.0.522.21309), X64 RyuJIT
  Job-FFFXRB : .NET 6.0.5 (6.0.522.21309), X64 RyuJIT

Runtime=.NET 6.0  IterationCount=20  WarmupCount=20

|              Method |      Mean |     Error |    StdDev | Ratio | RatioSD | Allocated |
|-------------------- |----------:|----------:|----------:|------:|--------:|----------:|
| StaticMethodExecute | 0.0878 ns | 0.0322 ns | 0.0344 ns |  1.00 |    0.00 |         - |
|     DelegateExecute | 1.1124 ns | 0.0263 ns | 0.0303 ns | 14.83 |    6.34 |         - |
|      PointerExecute | 0.9020 ns | 0.0447 ns | 0.0497 ns | 11.84 |    4.75 |         - |

@kamronbatman opened a pull request with the executed tests: [Pull Request](https://github.com/modernuo/ModernUO-Benchmarks/pull/3)

kamronbatman:
Might be better to emit IL in runtime to get the 0.08 speeds. But this looks like a little bit of a perf gain which is nice.

stefanomerotta:
> Might be better to emit IL in runtime to get the 0.08 speeds. But this looks like a little bit of a perf gain which is nice.

I run again the test with "no inline" the static method. I think that the code is somehow optimized and the static method was inlined.

|              Method |      Mean |     Error |    StdDev | Ratio | RatioSD | Allocated |
|-------------------- |----------:|----------:|----------:|------:|--------:|----------:|
| StaticMethodExecute | 0.6010 ns | 0.0550 ns | 0.0611 ns |  1.00 |    0.00 |         - |
|     DelegateExecute | 0.9125 ns | 0.0534 ns | 0.0593 ns |  1.53 |    0.20 |         - |
|      PointerExecute | 0.7191 ns | 0.0428 ns | 0.0458 ns |  1.21 |    0.14 |         - |

kamronbatman:
@stefanomerotta This looks great. What happens if we drop `managed` and just do `delegate*<>`. It should work, but how does that affect the benchmarks?

stefanomerotta:
> @stefanomerotta This looks great. What happens if we drop `managed` and just do `delegate*<>`. It should work, but how does that affect the benchmarks?

It's the same: without any keyword the default behavior is "managed"

stefanomerotta:
> So let's remove the word managed so it is less to type. The rest looks good, let's ship it!
> 
> Meanwhile my PR to remove the encoded packet handler will handle that use case. And I can handle the extended packet handler if yours doesn't.

If you want to check how I managed them it's better.
The "trick" that I used is the generic static class that are static per generic type, but this requires a specific mark for every extended packet handler, like "MapUOProtocolInfo": it works but maybe you don't like this approach.

We can refactor the extended handler system if this one doesn't fit the ModernUO code style.

kamronbatman:
This looks great!

Issue #1091, opened 2022-06-26
Add conditional support to [exportspawners


kamronbatman:
For example, it would be great to be able to do something like:`[facet exportspawners ./Spawns/yew.json where Name starts Yew_`

PR #1096, opened 2022-06-28
fix: Add optional conditional on exportspawners

markdwags:
Updated command to allow an optional conditional when exporting.

If I have 3 spawners, named: `Spawner_1`, `Yew_1`, and `Yew_2`, I could export just the spawners starting with `yew` via:

`[facet exportspawners Spawns\yew.json yew`

Ref: https://github.com/modernuo/ModernUO/issues/1091

### DIFF
--- a/Projects/UOContent/Engines/Spawners/Commands/ExportSpawnersCommand.cs
+++ b/Projects/UOContent/Engines/Spawners/Commands/ExportSpawnersCommand.cs
@@ -13,6 +13,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>. *
  *************************************************************************/
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using Server.Commands.Generic;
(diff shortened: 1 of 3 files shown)


Projects/UOContent/Engines/Spawners/Commands/ExportSpawnersCommand.cs
```diff
@@ -35,13 +36,14 @@ public ExportSpawnersCommand()
             Commands = new[] { "ExportSpawners" };
             ObjectTypes = ObjectTypes.Items;
             Usage = "ExportSpawners";
-            Description = "Exports the given the spawners to the a file";
+            Description = "Exports the given spawners to the a file";
```
kamronbatman, line 39: Not sure how my grammar got so bad. Can we change `the a file` to `a file`?

markdwags:
My use-case was after importing several large spawner files, I wanted to export per name to have more separation between the spawners and more granular control of the files I import/export.

It could be as simple or complex as required.  This is a simple implementation as you see, just matching the starting characters to the name of the spawner.

PR #1112, opened 2022-07-07
fix: Fixes curses and spell effects

mark1145:
* Fix statmod naming mismatch: everything in code searches for "[Magic] {type} Offset" but stat reductions are being added as "[Magic] {type} Curse"

### DIFF
--- a/Projects/UOContent/Engines/ConPVP/DuelContext.cs
+++ b/Projects/UOContent/Engines/ConPVP/DuelContext.cs
@@ -1817,12 +1817,18 @@ namespace Server.Engines.ConPVP
 
         public static void Debuff(Mobile mob)
         {
-            mob.RemoveStatMod("[Magic] Str Offset");
-            mob.RemoveStatMod("[Magic] Dex Offset");
-            mob.RemoveStatMod("[Magic] Int Offset");
+            mob.RemoveStatMod("[Magic] Str Buff");
+            mob.RemoveStatMod("[Magic] Dex Buff");
+            mob.RemoveStatMod("[Magic] Int Buff");
+            mob.RemoveStatMod("[Magic] Str Curse");
+            mob.RemoveStatMod("[Magic] Dex Curse");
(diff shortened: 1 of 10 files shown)


mark1145:
Hi @kamronbatman,

I found this small bug when playing around with remove curse

kamronbatman:
@mark1145, great catch.
We might need to do some research, but my understanding is that curses _are separate_ from buffs.
I tried to match ServUO, which supposedly fixed this. But I think it is broken there too.

To fix this, we should probably change all `Offset` to either `Buff` or `Curse` and do searches for `Curse` specifically.

mark1145:
Hi @kamronbatman,

The testing looks all good:
* Blood oath -> Tested removing the curse with a spell + allowing it to expire
* Weaken + Clumsy + Feeblemind -> Tested removing with a spell + allowing to expire (I noticed affect does not expire upon duration ending if they stand still, the character has to move for it to expire)
* Bless -> Tested offset works all okay. (I noticed affect does not expire upon duration ending if they stand still, the character has to move for it to expire)
* Curse -> Tested removing with a spell + allowing to expire
* Mind rot -> Tested removing the curse with a spell + allowing to expire
* Strangle -> Tested removing the curse with a spell + allowing to expire
* Spell plague -> Tested removing the curse with a spell. I could not see it expiring normally

I noticed a small bug with noble sacrifice: in original AOS it's supposed to also be the equivalent to a full remove curse, but the way the code is written, it will only remove 1 affliction and then nothing else due to the OR operators. I will fix it in a separate PR though
image

kamronbatman:
I was trying to be clever. Obviously a fail.
Let's break it out to something like:

```cs
sendEffect = RemoveEffect(m) || sendEffect;
```
And repeat that for each one.

mark1145:
Hi @kamronbatman, all done :)

kamronbatman (approved):
Yesss thank you!

PR #1135, opened 2022-08-01
* Weapons sometimes dropping a property

mark1145:
* Weapons sometimes dropping a property when a slayer group with no lesser slayers is rolled e.g. undead


mark1145:
Hi @kamronbatman,

I had a look at your suggestion to make separate lesser slayer lists if they are not used anywhere, it seems the lesser slayer lists are used even in pre-aos for a mechanic where people kill a particular group of mobs and they get a 5% chance to give them a slayer from that group.

As this issue seems specific to AOS+, I've gone for an approach to reroll the mod from a child list if they were not lucky on their 10% super slayer roll.

I've tested it and it seems all good.

Let me know if you want anything changed, thanks!

PR #1136, opened 2022-08-03
* Exceptional chance was always showing 55%

mark1145:


mark1145:
Hi @kamronbatman,

I found a minor bug, essentially the craft chance was always being set to 1 and so when the CraftItem.GetExceptionalChance() went to calculate exceptional chance, it would always be 1 - 0.45

kamronbatman:
@mark1145 I made a change. Let me know if this works too!

mark1145:
Hi @kamronbatman,

It looks good, I'm not sure why that if statement was there, I had thought maybe there was a narrow scenario when the person has exactly enough skill + uses a talisman but I can't see it does anything

kamronbatman:
> Hi @kamronbatman,
> 
> It looks good, I'm not sure why that if statement was there, I had thought maybe there was a narrow scenario when the person has exactly enough skill + uses a talisman but I can't see it does anything

Left over from long ago... Who knows 😀

PR #1138, opened 2022-08-17
* Fix players casting is sometimes not disturbed

mark1145:
Hi @kamronbatman,

I've fixed the fizzle bug and done some refactoring of the code to avoid duplicating. 

There is one thing extra I have changed as well which might seem irrelevant, but I think it should still be changed:
* Instead of checking if someone is casting by: using State == SpellState.Casting
* I use the Spell.IsCasting bool which underneath does exactly the same check: State == SpellState.Casting
* The reason is that this bool is what's used by other parts of the server e.g. dont let someone move if they are casting.
* In the future if someone does a custom change like letting people use spirit speak whilst holding a spell, they will want to make changes to the Spell.IsCasting bool so that the other parts are also checking if they are casting spirit speak

### DIFF
--- a/Projects/UOContent/Mobiles/Animals/Mounts/Ethereals.cs
+++ b/Projects/UOContent/Mobiles/Animals/Mounts/Ethereals.cs
@@ -386,15 +386,8 @@ namespace Server.Mobiles
                 Disturb(DisturbType.Hurt, false);
             }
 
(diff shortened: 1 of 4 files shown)


kamronbatman (approved):

kamronbatman:
Thank you!!

PR #1139, opened 2022-08-17
Focus Fixes

DocKnight86:
- Focus is a Pre-AOS skill.
- Utility.cs was off by one. Guessing this was just overlooked when Throwing skill was added.


kamronbatman:
Thank you for the catch!

PR #1140, opened 2022-08-18
fix: Solen teleporter deserialization null pointer exception

mark1145:
Hi @kamronbatman,

As we discussed sometime ago, the solen hiven teleporters cause null reference exceptions on server start up due to the setters having extra processing for updating their parent/child objects. The exception happens because there is no guarantee on world load that those objects will have been deserialised yet, therefore can be null.

I tested it and it works, let me know if anything needs changing. I wasn't sure on a name so I went with Raw.

Thanks!


kamronbatman:
This is great!

PR #1141, opened 2022-08-19
fix: Players abusing monster idle mode to kill them off-screen

mark1145:
Hi @kamronbatman,

This change fixes players being able to abuse monster idle mode to kill them without challenge by summoning EVs/blade spirits offscreen. If there are players etc. onscreen for the monster to target, it will still follow the old logic to target those first.

### DIFF
--- a/Projects/UOContent/Mobiles/AI/BaseAI.cs
+++ b/Projects/UOContent/Mobiles/AI/BaseAI.cs
@@ -2590,6 +2590,8 @@ public abstract class BaseAI
 
         Mobile newFocusMob = null;
         var val = double.MinValue;
+        Mobile enemySummonMob = null;
+        var enemySummonVal = double.MinValue;
 
         var eable = map.GetMobilesInRange(m_Mobile.Location, iRange);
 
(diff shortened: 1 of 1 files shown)


mark1145:
I forgot to mention this issue exists in AOS onwards

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2720,8 +2722,14 @@ public virtual bool AcquireFocusMob(int iRange, FightMode acqType, bool bPlayerO
             }
 
             var theirVal = m_Mobile.GetFightModeRanking(m, acqType, bPlayerOnly);
-
-            if (theirVal > val && m_Mobile.InLOS(m))
+            //The summon is targeted when nothing else around. Otherwise this monster enters idle mode, 
+            //which players can abuse by casting EVs offscreen and this monster wont fight back
+            if (Core.AOS && theirVal > enemySummonVal && m_Mobile.InLOS(m) && bc?.Summoned == true && bc?.Controlled != true)
```
kamronbatman, line 2727: Let's pull out the InLOS check so we don't have to potentially do it twice.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2730,7 +2738,7 @@ public virtual bool AcquireFocusMob(int iRange, FightMode acqType, bool bPlayerO
 
         eable.Free();
 
-        m_Mobile.FocusMob = newFocusMob;
+        m_Mobile.FocusMob = newFocusMob ?? enemySummonMob;
```
kamronbatman, line 2741: Looks like this variable is never set at the same time as `newFocusMob`, so we probably can reuse `newFocusMob` instead of `enemySummonMob`?

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2720,8 +2722,14 @@ public virtual bool AcquireFocusMob(int iRange, FightMode acqType, bool bPlayerO
...
 
             var theirVal = m_Mobile.GetFightModeRanking(m, acqType, bPlayerOnly);
-
-            if (theirVal > val && m_Mobile.InLOS(m))
+            //The summon is targeted when nothing else around. Otherwise this monster enters idle mode, 
+            //which players can abuse by casting EVs offscreen and this monster wont fight back
+            if (Core.AOS && theirVal > enemySummonVal && m_Mobile.InLOS(m) && bc?.Summoned == true && bc?.Controlled != true)
+            {
+                enemySummonMob = m;
+                enemySummonVal = theirVal;
```
kamronbatman, line 2730: This variable is never used.

PR #1144, opened 2022-08-22
spawn TreasureChestDirt next to treasure location, not ontop of it

Mink80:
Spawn TreasureChestDirt next to treasure location, not ontop of it. Else it will be found in the Map.CanFit() check and cancel the digging prcess with 503024 (You stop digging because something is directly on top of the treasure chest.)


kamronbatman:
After looking into this more, I think it's a bug with CanFit.

PR #1154, opened 2022-08-30
BaseAI: Fix wrong min/max delay for detect hidden NPC ability

stefanomerotta:
min/max were calculated with this formula:
```cs
var min = delay * (9 / 10); // 13s at 1000 int, 33s at 400 int, 54s at <250 int
var max = delay * (10 / 9); // 16s at 1000 int, 41s at 400 int, 66s at <250 int
```

Because of int type:
- (9 / 10) is always 0
- (10 / 9) is always 1

Changed to milliseconds and removed the need of TimeSpan conversion


kamronbatman:
Great find, thanks!

kamronbatman (approved):

PR #1165, opened 2022-09-06
feat: Implement Oil Flask & Refueling

### LINKED ISSUE #1079: Add OilFlask, EmptyOilFlask and refueling Lanterns
Add OilFlask, EmptyOilFlask.
Lanterns are refueled by dclicking the oil flask and targeting the lantern.
Add oil flasks to provisioner vendors and others that may sell them.

ricky26:
Fixes #1079

Added Empty Oil Flask, Oil Flask & Fish Oil Flask. Non-decorative lanterns will now run out of fuel.

Added recipe for Empty Oil Flask.

Added Oil Flask to the provisions supplier.

### DIFF
--- a/Projects/UOContent/Items/Lights/Lantern.cs
+++ b/Projects/UOContent/Items/Lights/Lantern.cs
@@ -6,7 +6,7 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
-    protected static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
+    public static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
 
     [Constructible]
     public Lantern() : base(0xA25)
--- a/Projects/UOContent/Items/Lights/Lantern.cs
+++ b/Projects/UOContent/Items/Lights/Lantern.cs
@@ -6,10 +6,12 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
+    public static readonly TimeSpan FullDuration = TimeSpan.FromMinutes(20);
+
     [Constructible]
     public Lantern() : base(0xA25)
     {
-        Duration = Burnout ? TimeSpan.FromMinutes(20) : TimeSpan.Zero;
+        Duration = FullDuration;
 
         Burning = false;
         Light = LightType.Circle300;
--- a/Projects/UOContent/Items/Lights/Lantern.cs
+++ b/Projects/UOContent/Items/Lights/Lantern.cs
@@ -6,7 +6,7 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
-    private static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
+    protected static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
 
     [Constructible]
     public Lantern() : base(0xA25)
--- a/Projects/UOContent/Items/Lights/Lantern.cs
+++ b/Projects/UOContent/Items/Lights/Lantern.cs
@@ -6,7 +6,7 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
-    public static readonly TimeSpan FullDuration = TimeSpan.FromMinutes(20);
+    private static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
 
     [Constructible]
     public Lantern() : base(0xA25)
--- a/Projects/UOContent/Items/Lights/Lantern.cs
+++ b/Projects/UOContent/Items/Lights/Lantern.cs
@@ -11,7 +11,7 @@ public partial class Lantern : BaseEquipableLight
     [Constructible]
     public Lantern() : base(0xA25)
     {
-        Duration = FullDuration;
+        Duration = Burnout ? FullDuration : TimeSpan.Zero;
 
         Burning = false;
         Light = LightType.Circle300;
--- a/Projects/UOContent/Items/Misc/OilFlask.cs
+++ b/Projects/UOContent/Items/Misc/OilFlask.cs
@@ -73,7 +73,9 @@ public partial class OilFlask : Item
                 eable.Free();
 
                 if (!didStack)
+                {
                     emptyFlask.MoveToWorld(Location, Map);
+                }
             }
 
             Consume(1);
--- a/Projects/UOContent/Items/Misc/OilFlask.cs
+++ b/Projects/UOContent/Items/Misc/OilFlask.cs
@@ -75,7 +75,7 @@ public partial class FishOilFlask : OilFlask
     }
 
     public override int LabelNumber => 1150863; // fish oil flask
-    public override int FilledMessageNumber => 1150865; //You fill the lamp with fish oil. It should burn for a nice long time.
+    public override int FilledMessageNumber => 1150865; // You fill the lamp with fish oil. It should burn for a nice long time.
     public override double FillMultiplier => 2;
 }
 
--- a/Projects/UOContent/Items/Misc/OilFlask.cs
+++ b/Projects/UOContent/Items/Misc/OilFlask.cs
@@ -1,0 +1,98 @@
+using ModernUO.Serialization;
+using Server.Targeting;
+
+namespace Server.Items;
+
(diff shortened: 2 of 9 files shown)


Projects/UOContent/Items/Lights/Lantern.cs
```diff
@@ -6,10 +6,12 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
+    public static readonly TimeSpan FullDuration = TimeSpan.FromMinutes(20);
+
     [Constructible]
     public Lantern() : base(0xA25)
     {
-        Duration = Burnout ? TimeSpan.FromMinutes(20) : TimeSpan.Zero;
+        Duration = FullDuration;
```
kamronbatman, line 14: Why change the conditional? I think the `Burnout` if I remember correctly is to have a configurable way for admins to change the way lighting works. 🤔

Projects/UOContent/Items/Misc/OilFlask.cs
```diff
@@ -0,0 +1,98 @@
...
+    public virtual int FilledMessageNumber => 1150864; // You fill the lamp with oil.
+    public virtual double FillMultiplier => 1;
+
+    public override void OnDoubleClick(Mobile from)
+    {
+        from.SendLocalizedMessage(501947); // Select lantern to refuel.
+        from.BeginTarget(2, true, TargetFlags.None, OnTargetPicked);
+    }
+
+    private void OnTargetPicked(Mobile from, object targeted)
```
kamronbatman, line 26: `BeginTarget<T>` exists, so let's use that instead. To do that, let's change targeted to `Item`

Projects/UOContent/Items/Misc/OilFlask.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        switch (targeted)
```
kamronbatman, line 28: While this works, let's use `if (targeted is Lantern lantern) {} else {}` instead.

Projects/UOContent/Items/Lights/Lantern.cs
ricky26: Ah okay, I assumed since it was hard-coded it was being used as a bookmark for things which don't burn out. I'll revert that.

Projects/UOContent/Items/Misc/OilFlask.cs
```diff
@@ -0,0 +1,98 @@
...
+                {
+                    lantern.Ignite();
+                }
+
+                from.SendLocalizedMessage(FilledMessageNumber);
+
+                var emptyFlask = new EmptyOilFlask();
+                if (Parent is Container container)
+                {
+                    container.AddItem(emptyFlask);
```
kamronbatman, line 52: Is this accurate? Most empty bottles and such are created in the user's backpack, regardless of the container. Let's check with drinking potions or something similar.
Usually we use `TryDropItem` so if the backpack is full, then we can put it on the ground.

Also to note: Empty flasks are stackable. So we may need code for dropping to the world that stacks with an existing flask stack, if it isn't full. We should look into how bouncing or other mechanics handle that and copy it.

Projects/UOContent/Items/Misc/OilFlask.cs
```diff
@@ -0,0 +1,98 @@
...
+[SerializationGenerator(0)]
+public partial class FishOilFlask : OilFlask
+{
+    [Constructible]
+    public FishOilFlask(int amount = 1) : base(amount)
+    {
+    }
+
+    public override int LabelNumber => 1150863; // fish oil flask
+    public override int FilledMessageNumber => 1150865; //You fill the lamp with fish oil. It should burn for a nice long time.
```
kamronbatman, line 78:
```suggestion
    public override int FilledMessageNumber => 1150865; // You fill the lamp with fish oil. It should burn for a nice long time.
```

kamronbatman (changes requested):
This looks great. Let's make the change outlined for consuming/creating the empty flask. Also please run `publish.cmd` so the migration files are generated!

Thank you!!!

Projects/UOContent/Items/Misc/OilFlask.cs
ricky26: The only instance of `BeginTarget<T>` I can see is one which takes `T state`, not a generic delegate.

kamronbatman: Oh I think most of the time `from.Target = new Target()`; is used.

kamronbatman: Per discussion, let's keep it the way it is for now.

Projects/UOContent/Items/Lights/Lantern.cs
```diff
@@ -6,10 +6,12 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
+    public static readonly TimeSpan FullDuration = TimeSpan.FromMinutes(20);
```
kamronbatman, line 9:
```suggestion
    private static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
```

Projects/UOContent/Items/Misc/OilFlask.cs
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            if (!from.PlaceInBackpack(emptyFlask))
+            {
+                var eable = from.GetItemsInRange(0);
+
+                if (!eable.Any(i => i.StackWith(from, i, false)))
+                {
+                    emptyFlask.MoveToWorld(Location, Map);
+                }
```
kamronbatman, line 67: Let's eliminate LINQ, it has terrible performance.
And I think StackWith should reference `this` as the argument, right?

```suggestion
                foreach (var i in eable)
                {
                    if (i.StackWith(from, this, false))
                    {
                        emptyFlask.MoveToWorld(Location, Map);
                        break;
                    }
                }
```

Projects/UOContent/Items/Lights/Lantern.cs
```diff
@@ -6,10 +6,12 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
+    private static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
```
kamronbatman, line 9:
```suggestion
    protected static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
```

ricky26: I think this change has broken the build. `FullDuration` is used by `OilFlask` when refueling.

Projects/UOContent/Items/Lights/Lantern.cs
```diff
@@ -6,10 +6,12 @@ namespace Server.Items;
 [SerializationGenerator(0, false)]
 public partial class Lantern : BaseEquipableLight
 {
+    protected static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
```
kamronbatman, line 9: I didn't realize it was used outside of Lantern.

```suggestion
    public static TimeSpan FullDuration = TimeSpan.FromMinutes(20);
```

Projects/UOContent/Items/Misc/OilFlask.cs
```diff
@@ -0,0 +1,117 @@
...
+                    {
+                        didStack = true;
+                        break;
+                    }
+                }
+
+                eable.Free();
+
+                if (!didStack)
+                    emptyFlask.MoveToWorld(Location, Map);
```
kamronbatman, line 76:
```suggestion
                if (!didStack)
                {
                    emptyFlask.MoveToWorld(Location, Map);
                }
```

Projects/UOContent/Items/Misc/OilFlask.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        from.SendLocalizedMessage(501947); // Select lantern to refuel.
+        from.BeginTarget(2, true, TargetFlags.None, OnTargetPicked);
+    }
+
+    private void OnTargetPicked(Mobile from, object targeted)
+    {
+        if (targeted is Lantern lantern)
+        {
+            if (!lantern.Movable || (lantern.Protected && from.AccessLevel == AccessLevel.Player))
```
kamronbatman, line 30:
```suggestion
            if (!lantern.Movable || lantern.Protected && from.AccessLevel == AccessLevel.Player)
```

kamronbatman (approved):

PR #1175, opened 2022-09-27
Fix abusing summons offscreen

mark1145:
Hey @kamronbatman,

Previously the monster AcquireFocusMob() would ignore player summons which was being abused by casting energy vortexes etc. out of range, the monster would not detect any players around and enter idle mode and die easily to the EVs; "player versus engine". 

With this change the monster still tries to acquire players but falls back to a player summon target if no valid player targets can be found.


kamronbatman:
@mark1145  This supersedes #1141?

mark1145:
Thanks I closed obsolete https://github.com/modernuo/ModernUO/pull/1141

Issue #1067, opened 2022-06-15
Add ISpanFormattable support to Geometry


kamronbatman:
Add ISpanFormattable to Point2D, Point3D etc. This will reduce the allocations while converting to a string by eliminating intermediate strings.

There are two ways to go about this:
1. Use the RawInterpolatedStringHandler and build it using `$"({X}, {Y}, {Z})"` or something similar and then get the raw buffer.
2. Use ValueStringBuilder and get the raw buffer.
3. Do it manually by creating static strings for each piece and filling in the holes with `X.TryFormat()` etc. This is the most typing, but true zero-allocation and probably the fastest.

We might need to get the length of the string that is being built, this will require getting the digit count, use this:
https://github.com/dotnet/runtime/blob/7f9ed8f271df8fb83862b3812d9c490e4ff74711/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/FormattingHelpers.CountDigits.cs#L95

PR #1203, opened 2022-10-24
Point2D implements ISpanFormattable

harleyholt:
Point2D implements a basic TryFormat function. Only a single format is supported: "(X, Y)" where X and Y are base 10 integers.

Uses recent improvements in string interpolation to write the characters to the destination without first allocating memory for boxed arguments or intermediate strings.

This is a partial solution to issue #1067. A similar solution well be implemented in other classes in Geometry if it looks promising.

Benchmarks
---
https://github.com/harleyholt/ModernUOBenchmarks/blob/d6f8dabf3c813937e65874ea82004701be0153c4/Program.cs

Before change:
```
BenchmarkDotNet=v0.13.2, OS=ubuntu 22.04
Intel Core i7-9700K CPU 3.60GHz (Coffee Lake), 1 CPU, 8 logical and 8 physical cores
.NET SDK=6.0.402
  [Host]     : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT AVX2


|                           Method |      Mean |    Error |   StdDev |   Gen0 | Allocated |
|--------------------------------- |----------:|---------:|---------:|-------:|----------:|
|                     CallToString |  44.27 ns | 0.068 ns | 0.063 ns | 0.0063 |      40 B |
|               InterpolatedString | 110.64 ns | 0.316 ns | 0.295 ns | 0.0126 |      80 B |
| InterpolatedStringMultiplePoints | 211.36 ns | 0.349 ns | 0.326 ns | 0.0293 |     184 B |
```

After change:
```
BenchmarkDotNet=v0.13.2, OS=ubuntu 22.04
Intel Core i7-9700K CPU 3.60GHz (Coffee Lake), 1 CPU, 8 logical and 8 physical cores
.NET SDK=6.0.402
  [Host]     : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT AVX2

|                           Method |      Mean |    Error |   StdDev |   Gen0 | Allocated |
|--------------------------------- |----------:|---------:|---------:|-------:|----------:|
|                     CallToString |  55.13 ns | 0.826 ns | 0.772 ns | 0.0063 |      40 B |
|               InterpolatedString |  55.73 ns | 0.954 ns | 0.892 ns | 0.0063 |      40 B |
| InterpolatedStringMultiplePoints | 105.42 ns | 0.151 ns | 0.134 ns | 0.0101 |      64 B |
```

### DIFF
--- a/Projects/Server/Geometry/Point2D.cs
+++ b/Projects/Server/Geometry/Point2D.cs
@@ -14,7 +14,6 @@
  *************************************************************************/
 
 using System;
-using Server.Buffers;
 
 namespace Server;
 
--- a/Projects/Server/Geometry/Point2D.cs
+++ b/Projects/Server/Geometry/Point2D.cs
@@ -14,6 +14,7 @@
  *************************************************************************/
 
 using System;
+using Server.Buffers;
 
 namespace Server;
 
@@ -117,7 +118,12 @@ public struct Point2D
     }
 
     public override string ToString() {
-        return $"{this}";
+        // Maximum number of characters that are needed to represent this:
+        // 4 characters for (, )
+        // Up to 11 characters to represent each integer
+        Span<char> span = stackalloc char[4+11*2];
+        this.TryFormat(span, out var charsWritten, null, null);
+        return span.Slice(0, charsWritten).ToString();
     }
 
     public string ToString(string format, IFormatProvider formatProvider)
@@ -126,4 +132,5 @@ public struct Point2D
         // default ToString implementation.
         return ToString();
     }
+
 }
--- a/Projects/Server/Geometry/Point2D.cs
+++ b/Projects/Server/Geometry/Point2D.cs
@@ -20,7 +20,7 @@ namespace Server;
 [Parsable]
 public struct Point2D
     : IPoint2D, IComparable<Point2D>, IComparable<IPoint2D>, IEquatable<object>, IEquatable<Point2D>,
-        IEquatable<IPoint2D>
+        IEquatable<IPoint2D>, ISpanFormattable
 {
     internal int m_X;
     internal int m_Y;
@@ -50,8 +50,6 @@ public struct Point2D
     public Point2D(Point2D p) : this(p.X, p.Y)
     {
     }
-
-    public override string ToString() => $"({m_X}, {m_Y})";
 
     public static Point2D Parse(string value)
     {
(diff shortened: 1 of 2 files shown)


Projects/Server/Geometry/Point2D.cs
```diff
@@ -112,4 +110,20 @@ public int CompareTo(IPoint2D other)
         var xComparison = m_X.CompareTo(other.X);
         return xComparison != 0 ? xComparison : m_Y.CompareTo(other.Y);
     }
+
+    public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan<char> format, IFormatProvider provider)
+    {
+        return destination.TryWrite(provider, $"({m_X}, {m_Y})", out charsWritten);
```
kamronbatman, line 116: Nice job! Not sure how I missed this. I'll use this for `Serial`!

Projects/Server/Geometry/Point2D.cs
```diff
@@ -112,4 +110,20 @@ public int CompareTo(IPoint2D other)
...
+    }
+
+    public override string ToString() {
+        return $"{this}";
```
kamronbatman, line 120: Is this better than what we had before with: `public override string ToString() => $"({m_X}, {m_Y})";`
I think it will end up being the same, or worse, incur an intermediate buffer before creating the final string.

kamronbatman (dismissed):

Projects/Server/Geometry/Point2D.cs
```diff
@@ -51,8 +51,6 @@ public Point2D(Point2D p) : this(p.X, p.Y)
     {
     }
 
-    public override string ToString() => $"({m_X}, {m_Y})";
```
kamronbatman, line 54: If we revert this, the benchmarks won't be slower I think?

harleyholt: Yeah, the longer execution time on this was a suprise. I'm trying to be clever to avoid competing, independent implementations of getting a string representation of this object (`ToString` vs interpolation, `$"{this}"`). I'd like to do a bit more exploration to understand better why there is a drop in performance and see if there is a solution to having one source of truth but keep the performance.

kamronbatman: Yeah I completely agree. I think the drop is in creating an intermediate ref struct and calling from one ref struct to another. It's only 8-10ns which makes sense.

Unfortunately, I don't think unlining will make this any better. I am ok with a 10ns penalty to keep the implementation standard. It only becomes a problem if you are doing some crazy like high perf text writing for a web service or file output with thousands of values.

kamronbatman: Nice! Let's do that. I should update Serial too.

Projects/Server/Geometry/Point2D.cs
```diff
@@ -112,4 +110,26 @@ public int CompareTo(IPoint2D other)
...
+
+    public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan<char> format, IFormatProvider provider)
+    {
+        return destination.TryWrite(provider, $"({m_X}, {m_Y})", out charsWritten);
+    }
+
+    public override string ToString() {
+        // Maximum number of characters that are needed to represent this:
+        // 4 characters for (, )
+        // Up to 11 characters to represent each integer
```
kamronbatman, line 122: How can 2 integers be 11 characters? 🤔

kamronbatman:
This looks great!

kamronbatman (approved):

PR #1208, opened 2022-10-25
fix: Adds AfterSerialize support. Removes BeforeSerialize support.

kamronbatman:
### Changes
* Implements an AfterSerialize method that is executed synchronously.
* Removes `BeforeSerialize` support since it was dangerous in its current implementation.
* Moves PlayerMobile kill/virtual decay to AfterSerialize.
* Adds kill decay to after Deserialize.

### DIFF
--- a/Projects/Server/Guild.cs
+++ b/Projects/Server/Guild.cs
@@ -39,6 +39,10 @@ public abstract class BaseGuild : ISerializable
     public abstract string Name { get; set; }
     public abstract GuildType Type { get; set; }
     public abstract bool Disbanded { get; }
+
+    public abstract bool ExecutesAfterSerialize { get; }
+    public abstract void AfterSerialize();
+
     public abstract void Delete();
 
     public bool Deleted => Disbanded;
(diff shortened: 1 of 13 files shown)


harleyholt:
I'm not able to provide much useful review of this as I don't yet understand the code base. I did wonder:

Will this change will cause problems for any works which depend on this? If so, is there some accepted way of messaging breaking changes?

It's not obvious to me what the original problem with `BeforeSerialize` was. Can you clarify for me?

kamronbatman:
> I'm not able to provide much useful review of this as I don't yet understand the code base. I did wonder:
> 
> Will this change will cause problems for any works which depend on this? If so, is there some accepted way of messaging breaking changes?
> 
> It's not obvious to me what the original problem with `BeforeSerialize` was. Can you clarify for me?

BeforeSerialize is new. I don't anticipate it has been used by anyone yet. With that said, most use cases can go into AfterSerialize. They are kind of interchangeable depending on how robust post deserialization is for that object.

Now why did I get rid of it. When I wrote it, I meant for it to be run before everything, and not in parallel. Unfortunately I didn't implement it correctly and the method is called in parallel which can be dangerous since game code should generally run on the game thread.

With that said, we could indiscriminately run BeforeSerialize and AfterSerialize on every object synchronously. It would slow down world saves though for larger servers. Think 12-26million objects interated twice in addition to the Parallel.ForEach.
I am still testing out this PR, but hoping that queuing objects that need to run a post Serialization will cut down on the amount of iterations when 99% of objects don't need it done.

PR #1231, opened 2022-11-06
fix: Adds ISpanFormattable to Geometry structs

### LINKED ISSUE #1067: Add ISpanFormattable support to Geometry

kamronbatman:
* Adds ISpanFormattable to geometry structs and makes ToString() near-zero-allocation.
* Adds IEquatable, and Parsable to Rectangle3D to get it in-line with the other structs.

Closes #1067

### DIFF
--- a/Projects/Server.Tests/Tests/Geometry/Point2DTests.cs
+++ b/Projects/Server.Tests/Tests/Geometry/Point2DTests.cs
@@ -1,17 +1,68 @@
 using System;
 using Xunit;
 
-namespace Server.Tests
+namespace Server.Tests;
+
+public class Point2DTests
 {
-    public class Point2DTests
+    [Fact]
+    public void TestPoint2DToString()
     {
-        [Fact]
-        public void TestToString()
-        {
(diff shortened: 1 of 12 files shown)


kamronbatman:
@harleyholt, up for reviewing? :)

PR #1254, opened 2022-11-19
fix: Fixes region priority and resolution.

kamronbatman:
- [X] Fixes world location JSON deserializer.
- [X] Fixes region resolver and priorities.
- [X] Removes DynamicJson from regions.
- [X] Adds missing region music.


Projects/Server/Regions/Region.cs
```diff
@@ -131,54 +172,6 @@ public Region(string name, Map map, Region parent, params Rectangle3D[] area)
...
-        if (Parent == null)
-        {
-            ChildLevel = 0;
-            Priority = DefaultPriority;
-        }
-        else
-        {
-            ChildLevel = Parent.ChildLevel + 1;
-            Priority = Parent.Priority;
-        }
```
kamronbatman, line 151: This logic is missing from the new JsonDto.

PR #1259, opened 2022-11-21
fix: check for same map in BaseCreature.TeleportPets()

Mink80:


Projects/UOContent/Mobiles/BaseCreature.cs
```diff
@@ -3841,7 +3841,7 @@ public static void TeleportPets(Mobile master, Point3D loc, Map map, bool onlyBo
             var eable = master.GetMobilesInRange(3);
             foreach (var m in eable)
             {
-                if (m is BaseCreature
+                if (m.Map == master.Map && m is BaseCreature
```
kamronbatman, line 3844:
```suggestion
                if (m is BaseCreature
```

kamronbatman: Revert this since `GetMobilesInRange` is done by `master.Map.GetMobilesInRange` which is constrained by that map.

kamronbatman:
Thank you!

PR #1337, opened 2023-02-09
Moved a few packets from Server to UOContent project

paezao:
This is part of the effort to better separate the Core Server from actual UO code.


Projects/Server/Network/NetState/NetState.cs
```diff
@@ -70,8 +67,8 @@ public partial class NetState : IComparable<NetState>
     private readonly long[] _packetCounts = new long[0x100];
     private string _disconnectReason = string.Empty;
 
-    internal int _authId;
-    internal int _seed;
+    public int authId;
```
kamronbatman, line 70: Let's use properties (getters and setters)

Projects/Server/Network/NetState/NetState.cs
```diff
@@ -70,8 +67,8 @@ public partial class NetState : IComparable<NetState>
...
+    public int seed;
```
kamronbatman, line 71: Same as above

PR #1392, opened 2023-04-15
fix: Fixes doors as decorations & optimizes them

### LINKED ISSUE #1391: Decorate Command Portcullis
The deco file crashes the server if there is a PortcullisNS or PortcullisEW in the  file when you use the decorate command but if you add 
```
[Constructible]
        public PortcullisNS(DoorFacing facing) : base(0x6F5, 0x6F5, 0xF0, 0xEF, new Point3D(0, 0, 20))
        {
        }
```

and
```
[Constructible]
        public PortcullisEW(DoorFacing facing) : base(0x6F6, 0x6F6, 0xF0, 0xEF, new Point3D(0, 0, 20))
        {
        }
```
 
to the Portcullis.cs script, it will generate fine when using the decorate command

kamronbatman:
### Summary
- [X] Fixes old RunUO bug with Portcullis causing crashing when used as a decoration.
- [X] Optimizes doors so they don't use a timer object unless they need one.
- [X] Converts doors to codegen
- [X] Optimized decoration command so it doesn't create extra strings/arrays when it doesn't need to.

Closes #1391


Distribution/Data/Decoration/Britannia/jhelom.cfg
```diff
@@ -1,5 +1,5 @@
 # portcullis
-Static 0x06F5
+PortcullisNS 0x06F5
```
kamronbatman, line 2: Revert this.

kamronbatman:
Gotta update decorations that set ItemID when they shouldn't.

PR #1395, opened 2023-04-18
fix: Ambiguous AddRes signatures causing bad necro regs display

mark1145:


Projects/UOContent/Engines/Craft/DefInscription.cs
```diff
@@ -215,7 +215,7 @@ private void AddNecroSpell(int spell, int mana, double minSkill, Type type, para
 
         for (var i = 1; i < regs.Length; ++i)
         {
-            AddRes(index, regs[i], 1, 501627);
+            AddRes(index, regs[i], null, 1);
```
kamronbatman, line 218: I think instead, the args are backwards?

kamronbatman:
Looks like am ambiguous set of AddRes signatures. I'll fix it. Thanks!

Projects/UOContent/Engines/Craft/DefInscription.cs
kamronbatman, line 218:
```suggestion
            AddRes(index, regs[i], null, 1, 501627);
```

kamronbatman (approved):

PR #1413, opened 2023-06-20
Fix mana drain debug build crash

vexyl:
Crashes on debug build when a player or mobile casts a mana drain on a player that has 0 mana and doesn't resist it. ManaDrain.cs line 71 uses Utility.Random(1, m.Mana), but there's a debug assert  in BaseRandomSource.cs that makes sure the count parameter (m.Mana in this case) is not 0.


kamronbatman:
Nice find and fix, thank you! I think we should add an if check for Mana <= 0 though. I can make that change.

vexyl:
No problem. We ran into on Sunny's sever, glad to be of help.

kamronbatman (approved):

PR #1420, opened 2023-07-04
Prevent setting properties of higher privileged mobiles

Mink80:


Projects/UOContent/Commands/Properties.cs
```diff
@@ -363,6 +363,13 @@ public static string SetValue(Mobile from, object o, string name, string value)
         {
             try
             {
+                if (obj is Mobile target &&
+                    (target.AccessLevel > from.AccessLevel ||
+                     target.Account is {} acc && acc.AccessLevel > from.AccessLevel))
+                {
+                    return "You can not modify a higher privileged mobile.";
+                }
```
kamronbatman, line 371: While I agree we need this, it should already be handled... I'll look around and see if the checks got borked.

Mink80: It's not. I tested it.
Also: this check prevents the modification of a staff character that is currently on a lower AccessLevel because of a [staffaccess usage

kamronbatman: Good call on the staff access, thanks!

kamronbatman (approved):

PR #1421, opened 2023-07-04
Exclude higher privileged mobiles in [admin gump lists

Mink80:


Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -922,6 +934,9 @@ public class AdminGump : Gump
                         AddLabel(20, 250, LabelHue, "Character Count:");
                         AddLabel(200, 250, LabelHue, charCount.ToString());
 
+                        AddLabel(250, 250, LabelHue, "Max Houses:");
```
kamronbatman, line 937: This doesn't exist.

PR #1422, opened 2023-07-04
Fixes GetAllSharedAccounts

Mink80:
The method had a non functional logic and crashed. This patch fixes both.

### DIFF
--- a/Projects/UOContent/Gumps/AdminGump.cs
+++ b/Projects/UOContent/Gumps/AdminGump.cs
@@ -6,7 +6,6 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using Server.Accounting;
 using Server.Buffers;
-using Server.Collections;
 using Server.Commands;
 using Server.Misc;
 using Server.Multis;
--- a/Projects/UOContent/Gumps/AdminGump.cs
+++ b/Projects/UOContent/Gumps/AdminGump.cs
@@ -6,6 +6,7 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using Server.Accounting;
 using Server.Buffers;
+using Server.Collections;
 using Server.Commands;
 using Server.Misc;
 using Server.Multis;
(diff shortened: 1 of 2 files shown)


Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -1579,30 +1580,41 @@ public void AddFirewallHeader()
 
                 for (var i = 0; i < theirAddresses.Length; ++i)
                 {
-                    table.TryAdd(theirAddresses[i], new List<Account> { acct });
+                    // if its an ip that is not in the list, add it
+                    if (!table.TryAdd(theirAddresses[i], new List<Account> { acct }))
```
kamronbatman, line 1584: We should do the opposite since this will needlessly create a new list every time.

Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -1579,30 +1580,41 @@ public void AddFirewallHeader()
...
+                    {
+                        // else, add the account to the existing ip
+                        table[theirAddresses[i]].Add(acct);
+                    }
                 }
             }
 
-            var tableEntries = table.ToList();
+            // Create a separate list to store items to be removed
+            using var itemsToRemove = PooledRefList<IPAddress>.Create();
```
kamronbatman, line 1593: Nice!

Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -1579,30 +1580,41 @@ public void AddFirewallHeader()
...
+                    itemsToRemove.Add(kvp.Key);
                 }
             }
 
+            // remove all entries that have only one account from the table
+            foreach (var ip in itemsToRemove)
+            {
+                table.Remove(ip);
+            }
+            itemsToRemove.Dispose();
```
kamronbatman, line 1610: We don't need the dispose if there is a `using`

Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -1579,30 +1580,41 @@ public void AddFirewallHeader()
...
+
+            // sort the table by the number of accounts per ip
+            var tableEntries = table.ToList();
             tableEntries.Sort(SharedAccountComparer.Instance);
 
-            return tableEntries;
+            // return the reversed list so that the highest number of accounts per ip is first
+            return Enumerable.Reverse(tableEntries).ToList();
```
kamronbatman, line 1617: We can actually do the opposite, add the items we want to the pooled array, then use a reverse iterator.

kamronbatman (approved):

PR #1425, opened 2023-07-15
feat: Customize expansion and set maps on first boot

mdodkins:
* Modifed ServerConfigurationPrompts to prompt for maps to be excluded. Default is inclusion. Only maps for the expansion selected are displayed
* Created MapSelection and tests for it, to define and format which maps have been selected
* Added SelectedMaps to ServerSettings
* Modified PublicMoongate so that only moongates for selected maps are generated with the world building command
* Modified moongate gump so that destinations presented are only from maps that have been selected
* Modified world building so that only spawn for selected maps is generated
* Updated CharacterCreation to make some reasonable guesses about where to spawn characters, based on maps selected (not great to appear in a map that has no moongates or spawn, for example)


Projects/Server/Maps/MapSelection.cs
```diff
@@ -0,0 +1,117 @@
+using System;
+using System.Collections.Generic;
+
+namespace Server.Maps
+{
+    public class MapSelection
+    {
+        public MapSelection() { }
```
kamronbatman, line 8: Duplicative, can be deleted.

Projects/Server/ExpansionInfo.cs
```diff
@@ -188,21 +192,89 @@ public static string GetEraFolder(string parentDirectory)
...
+    }
+
+    public static void LoadConfiguration(Expansion currentExpansion)
+    {
+        // As we are replacing an existing expansion configuration, we need to make sure
+        // that the default ones have been loaded
+        if (expansionConfigs == null)
+        {
+            Console.WriteLine("Failed to load expansion.json due to expansion configs not being loaded");
+            return;
```
mdodkins, line 238: maybe I should just throw an exception here instead....

kamronbatman: If they deleted expansion.json, we should just generate one from modernuo.json/expansions.json.

If they are missing expansions.json, then throw. And all of this logic should be executed _after_ modernuo.json is generated so we don't have an order of operations issue.

mdodkins: Ah yep, cool, I'll do that 🙂

Projects/Server.Tests/Tests/Maps/MapSelectionTests.cs
```diff
@@ -0,0 +1,138 @@
...
+using Server.Json;
+using System.Text.Json.Serialization;
+using System.Text.Json;
+
+namespace Server.Tests.Tests.Maps
+{
+    public class MapSelectionTests
+    {
+        [Fact]
+        public void No_maps_enabled_on_construction()
```
kamronbatman, line 13: I don't have all of the rules on (analyzers are hard to configure properly, lots of conflicting stuff).
But it would definitely flag these function names as not being part of the style.
Let's use something like:
```cs
public void TestNoMapsEnabled()
{
}
```

kamronbatman (changes requested):

Projects/Server/Maps/MapSelection.cs
```diff
@@ -0,0 +1,119 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Server.Maps;
+
+public class MapSelection
+{
+    private MapSelectionFlags _selectedMaps;
+    private static readonly Array _mapSelectionValues = Enum.GetValues(typeof(MapSelectionFlags));
```
kamronbatman, line 10: Let's avoid `Array` because it is a legacy type. Instead use `MapSelectionFlags[]` so it forces the compiler to use `Enum.GetValues<T>()` :)

kamronbatman (changes requested):

Projects/Server/Maps/MapSelection.cs
```diff
@@ -0,0 +1,119 @@
...
+
+    public bool Includes(MapSelectionFlags mapFlags)
+    {
+        if ((_selectedMaps & mapFlags) == mapFlags)
+        {
+            return true;
+        }
+
+        return false;
+    }
```
kamronbatman, line 44:
```suggestion
    public bool Includes(MapSelectionFlags mapFlags) => (_selectedMaps & mapFlags) == mapFlags;
```

Projects/Server/Maps/MapSelection.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                continue;
+            }
+
+            if (value.ToString() != mapName)
+            {
+                continue;
+            }
+
+            return (_selectedMaps & value) == value;
```
kamronbatman, line 60:
```suggestion
            return Includes(value);
```

Projects/Server/Maps/MapSelection.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+        formatted = formatted.TrimEnd().Replace(" ", ", ");
+
+        if (formatted == "")
+        {
+            formatted = "None";
+        }
+
+        return formatted;
+    }
```
kamronbatman, line 118:
```suggestion
    public string ToCommaDelimitedString()
    {
        using var builder = ValueStringBuilder.Create();

        for (var i = 0; i < _mapSelectionValues.Length; i++)
        {
            var value = _mapSelectionValues[i];

            if (value > 0 && Includes(value))
            {
                  if (i < _mapSelectionValues.Length - 1)
                  {
                      builder.Append($"{value}, ");
                  }
                  else
                  {
                      builder.Append($"{value}");
                  }
            }
        }

        return builder.Length == 0 ? "None" : builder.ToString();
    }
```

Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -2091,23 +2092,30 @@ public override void OnResponse(NetState sender, RelayInfo info)
                                     }
                                     else
                                     {
-                                        if (Core.ML)
+                                        if (Core.ML && Core.AvailableMaps.Includes(MapSelectionFlags.Malas))
```
kamronbatman, line 2095: I am thinking it might make more sense to move it from `Core` to `Map`?

Projects/UOContent/Misc/CharacterCreation.cs
```diff
@@ -256,11 +258,22 @@ private static CityInfo GetStartLocation(CharacterCreatedEventArgs args, bool is
...
+                    break;
                 }
             case "samurai":
                 {
-                    if ((flags & ClientFlags.Tokuno) != 0)
+                    bool haotisAndTokunoAccessible = 
+                        ((flags & ClientFlags.Tokuno) == ClientFlags.Tokuno) &&
+                        ((flags & ClientFlags.Malas) == ClientFlags.Malas) &&
+                        Core.AvailableMaps.Includes(MapSelectionFlags.Malas) &&
+                        Core.AvailableMaps.Includes(MapSelectionFlags.Tokuno);
```
kamronbatman, line 274: These can be combined:

```suggestion
                        Core.AvailableMaps.Includes(MapSelectionFlags.Malas | MapSelectionFlags.Tokuno)
```

Projects/UOContent/Misc/CharacterCreation.cs
```diff
@@ -280,7 +293,13 @@ private static CityInfo GetStartLocation(CharacterCreatedEventArgs args, bool is
                 }
             case "ninja":
                 {
-                    if ((flags & ClientFlags.Tokuno) != 0)
+                    bool enimosAndTokunoAccessible =
+                        ((flags & ClientFlags.Tokuno) == ClientFlags.Tokuno) &&
+                        ((flags & ClientFlags.Malas) == ClientFlags.Malas) &&
+                        Core.AvailableMaps.Includes(MapSelectionFlags.Malas) &&
+                        Core.AvailableMaps.Includes(MapSelectionFlags.Tokuno);
```
kamronbatman, line 300: Same

Projects/Server.Tests/Tests/Maps/MapSelectionTests.cs
mdodkins: sure :+1:

Projects/Server/Maps/MapSelection.cs
mdodkins: ooo alright! :)

Projects/Server/Maps/MapSelection.cs
mdodkins: ah yes, still getting used to things like that haha

Projects/UOContent/Gumps/AdminGump.cs
mdodkins: seems reasonable!

mdodkins:
thanks for the review!

Projects/Server/Maps/MapSelection.cs
mdodkins: ahh, Enum.GetValues(typeof(MapSelectionFlags)) returns an Array, but Enum.GetValues<MapSelectionFlags>() returns MapSelectionFlags, cool :+1:

mdodkins:
OK @kamronbatman I think I've resolved all that, cheers!

.gitignore
```diff
@@ -25,6 +26,7 @@
 /Projects/*/bin
 /Projects/*/Generated
 
+*.swp
```
kamronbatman, line 29: swp file?

kamronbatman (approved):

kamronbatman:
I did some preliminary cleanup and simplification. Also removed ExpansionConfig by combining it with ExpansionInfo.

Projects/Server/Maps/Map.cs
```diff
@@ -320,7 +320,6 @@ public sealed class Map : IComparable<Map>, ISpanFormattable, ISpanParsable<Map>
     public const int SectorActiveRange = 2;
 
     private static ILogger logger = LogFactory.GetLogger(typeof(Map));
-
```
kamronbatman, line 323: We can revert this so we have a smaller diff.

PR #1436, opened 2023-07-29
fix: Fixes untamed creatures not entering houses

vexyl:
On Dark Factions, untamed creatures were unable to move into houses (e.g., if attacked and lured). I noticed 2 places where the Controlled flag was checked in HouseRegion.cs OnMoveInto(). They both seem to be doing the same thing except one checks if AoSRules and if it's public. I'm guessing the latter is the one wanted. This removes the first check.


kamronbatman:
I think this worked on RunUO, so I am going to check the code to see what I changed first.

kamronbatman (approved):

PR #1443, opened 2023-08-08
Container enumeration of all nested content

stefanomerotta:
First try of Container enumeration "GetAllItems" with ref struct enumerable and LINQ-Like methods.

__This is a proposal: this PR have to be deeply reviewed before the merge.__

@kamronbatman any opinion about this?


Projects/Server/Items/Container.Enumerable.cs
```diff
@@ -0,0 +1,226 @@
...
+using Server.Common.Filters;
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+
+namespace Server.Items;
+
+public partial class Container
+{
+    private static readonly Queue<Container> containersForGetItem = new();
```
kamronbatman, line 11: We can use `PooledRefQueue` instead.

Projects/Server/Items/Container.Enumerable.cs
```diff
@@ -0,0 +1,226 @@
...
+        public AllItemsEnumerable<ChainFilter<Item, TResult, TResult, TFilter, TakeFilter<TResult>>, TResult> Take(int amount)
+        {
+            return new(container, new(filter, new(amount)));
+        }
+
+        public TResult? FirstOrDefault(TResult? @default = default)
+        {
+            if (container.m_Items is not { Count: > 0 })
+            {
+                return @default;
```
kamronbatman, line 60:
```suggestion
                return default;
```

Projects/Server/Items/Container.Enumerable.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            return new(container, new(filter, default));
+        }
+
+        public AllItemsEnumerable<ChainFilter<Item, TResult, TResult, TFilter, TakeFilter<TResult>>, TResult> Take(int amount)
+        {
+            return new(container, new(filter, new(amount)));
+        }
+
+        public TResult? FirstOrDefault(TResult? @default = default)
```
kamronbatman, line 56: There are libraries that will codegen this for us.

Projects/Server/Common/Filters/IFilter.cs
```diff
@@ -0,0 +1,6 @@
+namespace Server.Common.Filters;
+
+public interface IFilter<in TIn, out TOut>
```
kamronbatman, line 3: Can we use an `Action<>` directly?

Projects/Server/Items/Container.Enumerable.cs
```diff
@@ -0,0 +1,226 @@
...
+                toRet.Add(item);
+            }
+
+            return toRet;
+        }
+
+        public TestAllItemsEnumerator<TFilter, TResult> GetEnumerator() => new(container, in filter);
+    }
+
+    public ref struct TestAllItemsEnumerator<TFilter, TResult>
```
kamronbatman, line 121: Technically Items can have `m_Items`, so this can be put generalized even more.
With that said, we can put this in it's own file I think?

Projects/Server/Items/Container.Enumerable.cs
```diff
@@ -0,0 +1,226 @@
...
+            containersForGetItem.Clear();
+
+            return @default;
+        }
+
+        public bool Any() => FirstOrDefault() != null;
+
+        public PooledRefQueue<TResult> ToPooledRefQueue()
+        {
+            PooledRefQueue<TResult> toRet = new();
```
kamronbatman, line 96: We don't support `new()` for this ref queue. Change to `PooledRefQueue.Create()`

Projects/Server/Common/Filters/IFilter.cs
stefanomerotta: If we use a function directly we will incur in heap allocations. With the struct filters we can implements filters even with closures (that cannot be cached by default) without allocations.
I made the "PredicateFilter" for fallback to normal Predicate or Func if needed by developers.
I know that create an ad-hoc struct for every case where we need a filter will produces many boilerplate code than a normal function creates a boilerplate code.
We can even switch to Func but the overhead of concatenation could be noticeable
.
As you suggested maybe it's better to switch to a dedicated library for LINQ source-generated when needed, that is surely more optimized and complete than my implementation.

Projects/Server/Items/Container.Enumerable.cs
```diff
@@ -0,0 +1,226 @@
...
+            return false;
+        }
+
+        public readonly void Dispose()
+        {
+            containers.Dispose();
+        }
+    }
+
+    public ref struct TestItemsEnumerator<TFilter, TResult>
```
kamronbatman, line 189: Looks like the difference between TestAllItems and TestItems is recursiveness. Should we use a flag instead of two implementations?

Projects/Server/Items/Container.Enumerable.cs
stefanomerotta: Yep, this static queue is used only for "FirstOrDefault": I tried with a static queue because in that case we don't need a temporary queue that survive outside the single method execution and maybe a standard queue (in this single thread code execution) could be better than ask from the pool a queue for every execution.
But I can switch to the pooledrefqueue: in this case the implementation could be easier because it's like "call the enumerator and do MoveNext one time".

Projects/Server/Items/Container.Enumerable.cs
stefanomerotta: Yes there are similarities between Item and Container.
I can switch to "Items" property instead of m_Items, that is the only common accessor for both classes, so the code can be used with little modifications also for standard Item class if required.
I don't know the use cases of m_Items for standard Item class btw.

Projects/Server/Items/Container.Enumerable.cs
stefanomerotta: My bad, I rewrote those code too many times that I forgot to use the Create method 😄

Projects/Server/Common/Filters/Filters.cs
```diff
@@ -0,0 +1,84 @@
...
+
+public readonly struct DummyFilter<T> : IFilter<T, T>
+{
+    public T? Invoke(T instance)
+    {
+        return instance;
+    }
+}
+
+public readonly struct ChainFilter<TIn, TMiddle, TOut, TFirst, TSecond> : IFilter<TIn, TOut>
```
kamronbatman, line 13: I love the idea of struct filters. Let's do two things:

1. Move the filters to their own set of files since they should be generic enough to be used outside of just containers.
2. Let's benchmark this. We have a benchmark repo (which probably needs updating, I can do that later). I would love to test out multiple approaches.

Projects/Server/Common/Filters/Filters.cs
```diff
@@ -0,0 +1,84 @@
+using System;
+
+namespace Server.Common.Filters;
+
+public readonly struct DummyFilter<T> : IFilter<T, T>
```
kamronbatman, line 5: Let's rename this to `IdentityFilter`

Projects/Server/Common/Filters/Filters.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    public TOut? Invoke(TIn instance)
+    {
+        return first.Invoke(instance) is not { } middle ? default : second.Invoke(middle);
+    }
+}
+
+public readonly struct PredicateFilter<T> : IFilter<T, T>
+{
+    private readonly Predicate<T> predicate;
```
kamronbatman, line 34: We might want to do specific benchmarking on this. I wonder about JIT optimization here since I think even using a static `Predicate<T>` might be really slow to execute. If we don't have lots of use cases, maybe someone should create a specific predicate-style filter for this like we do with Comparers.

Projects/Server/Items/Container.Enumerable.cs
stefanomerotta: Another miss...the name of this iterator is wrong 😄 
The implementation for recursive and non recursive is quite different: non-recursive don't need queues as buffer and have less code statements, so I thought to separate entirely both and expose the non-recursive on a different set.

If you want I still can introduce the flag as the standard RunUO implementation and use a single path for both cases

kamronbatman:
This is a great start. Lots of this work reminds me of HyperLinq and StructLinq. I like their LINQ-style builder pattern better than nested struct construction. I wonder if there is a library like one of those that is currently maintained.

Projects/Server/Items/Container.Enumerable.cs
stefanomerotta: Should be fixed now

stefanomerotta:
First basic benchmark for Get/Find items by type where type is just "Item" (aka get AllItems from nested containers).

```
|                            Method |     Mean |     Error |    StdDev |  Allocated |
|---------------------------------- |---------:|----------:|----------:|-----------:|
|          RecursiveFindAllItemsOld | 8.966 ms | 0.1757 ms | 0.3169 ms | 8255.88 KB |
|             RecursiveFindAllItems | 7.856 ms | 0.1533 ms | 0.2386 ms | 8194.26 KB |
|          RecursiveFindAllItemsNew | 9.218 ms | 0.1818 ms | 0.2233 ms | 8257.75 KB |
| RecursiveFindAllItemsNewIteration | 3.819 ms | 0.0756 ms | 0.0840 ms |   63.94 KB |
```

RecursiveFIndAllItemsOld: is the actual implementation with BFS from ModernUO, that btw have ha bug
RecursiveFindAllItems: is an implementation with list like the standard one but with DFS, so no temporary queue but recursion
RecursiveFinAllItemsNew: is the enumerable from this PR that uses the "ToList" to materialize the collection like the other two previous methods
RecursiveFinAllItemsNewIteration: is the same enumerable method but with a foreach instead of materialized list, that should be used when collection will not be modified while iterated (I simulated a TotalCount of retrieved items).

kamronbatman:
I think we handled this for the most part. We can revisit filters and other advanced features another time if we get a use-case for it.

PR #1451, opened 2023-08-11
fix: Player vendor gump update so it can construct items with single parameters

mdodkins:
Player vendor customization totally broken without this...


kamronbatman:
Thank you!!!

kamronbatman (approved):

PR #1454, opened 2023-08-12
Replaced FindItem(s)ByType(s) implementation with BFS strategy

stefanomerotta:
Replaced FindItem(s)ByType(s) implementation with BFS strategy.
Some minor changes due to different output collection.


kamronbatman:
@stefanomerotta - I updated the CLA (nothing really material), so please resign.

kamronbatman:
Thoughts on using `PooledRefList` instead of `List<item>` for some of the internal uses?

PR #1466, opened 2023-08-21
fix: Fixes JSON TypeConverter so that it can deserialize full name types

mdodkins:
TypeConverter wasn't checking for fully qualified type names, which meant it couldn't convert type names into types that have been serialized by it.

Fixed this and added a test. Would be possible to expand this test into a full suite of conversion tests...


kamronbatman:
Thank you for also adding tests.

kamronbatman (approved):

PR #1471, opened 2023-08-25
fix: Fixes expansions.json to correctly reflect name of supported feature flags

mdodkins:
When the expansions.json file is deserialized to ExpansionInfo[], it has been clearing the feature flags. This is because the json deserializer has been looking for "FeatureFlags" but the name in the json is "SupportedFeatures". I've corrected expansions.json to match this, and verified it now works.


kamronbatman:
I am not sure how I screwed that up so badly... wow

kamronbatman (approved):

PR #1481, opened 2023-08-31
Added docker configuration

thecatontheflat:


thecatontheflat:
Not much in the server logs. Just acknowledging a client was connected, and then disconnected (once I selected the shard in the game client):

image

kamronbatman:
No it doesn't. I have never seen that. I would need to dig through the code too in order to find out.

thecatontheflat:
https://github.com/ServUO/ServUO/pull/5047/files

They had a similar issue. Will port the fix here too

thecatontheflat:
Essentially, to make it a smooth developer experience and to containerize it properly, these things need to be addressed:

- Server IP binding when connecting from the host machine (can be either solved via proposed change to the `ServerList`, or by offloading this configuration to the developer and let them figure it out with Docker (or Docker Compose) networks.
- Exposing server configuration via mount or via environmental variables. Currently, to change server configuration, it needs to be changed within the container.
- For an easier development process, the project files can be mounted from the host via volume binds, in which case it requires mapping user from host to container

It felt like it started bringing more issues than fixes, so I didn't want to steal anyone's attention.

Thoughts?

kamronbatman:
Can we address this by modifying docker-compose to include a network bridge?

PR #1487, opened 2023-09-03
Starting a server notes and tips.

BCrowly:
Just a note and tip added to docs.

### DIFF
--- a/docs/starting-server.md
+++ b/docs/starting-server.md
@@ -20,3 +20,11 @@ Everything is run from the _Distribution_ folder.
     cd Distribution
     dotnet run ModernUO.dll
     ```
+
+=== "Game Files"
+    !!! Note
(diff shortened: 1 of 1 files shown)


docs/starting-server.md
```diff
@@ -20,3 +20,11 @@ Everything is run from the _Distribution_ folder.
...
     dotnet run ModernUO.dll
     ```
+
+=== "Game Files"
+    !!! Note
+        Game files are required to initiate a server. Upon the initial launch, you will encounter a prompt asking for the location of your game files or ClassicUO installation. ClassicUO installations should be configured to reference game files directory within settings.json file.
+
+
+    !!! Tip
+        Game files are not directly distributed in ModernUO, find the latest version to download these at [client download](https://downloads.eamythic.com/uo/installers/UOClassicSetup_7_0_24_0.exe)
```
thecatontheflat, line 30: https://uo.com/Client-Download/

Perhaps this would be a better URL (so it's not locked to a specific version)?

BCrowly: Maybe that's better than direct download indeed.

BCrowly:
Updated link suggestion.

Issue #1486, opened 2023-09-03
Configurable skill/stat gain factor?


thecatontheflat:
I want to run a test server with accelerated skills and stats gain.
Currently I have 2 options when it comes to skill gain:

- Either modify it in the [skills.json](https://github.com/modernuo/ModernUO/blob/main/Distribution/Data/skills.json#L13) for each skill
- Or modify it directly in the [SkillCheck.cs](https://github.com/modernuo/ModernUO/blob/main/Projects/UOContent/Skills/SkillCheck.cs#L109) logic

For a better modularity, I thought it would make sense to make this configurable.

Would you accept a PR with such a change?

kamronbatman:
Yeah I agree.

thecatontheflat:
@kamronbatman While refactoring this, I figured that some skills should allow for multiple stat gains, e.g.:
https://github.com/modernuo/ModernUO/blob/main/Distribution/Data/skills.json#L24-L26

While in the current implementation it seems that Str always takes priority and the rest of the logic is not executed if Str increase rolls a success:
https://github.com/modernuo/ModernUO/blob/main/Projects/UOContent/Skills/SkillCheck.cs#L245-L256

Is this intended behavior? If not, I'll change it to separate "if" statements so they all get evaluated (unless `XXXGain` is 0)

kamronbatman:
Note that they are not `else-if` and it does not return, So it actually tries to gain each one, every time.

thecatontheflat:
Could you please help me understand it?

From what I see, the logic is triggered once - e.g. `PlayerMobile` succeeded at using `Meditation`, which then triggers `m.CheckSkill`, which then triggers `SkillCheckLocationHandler` etc. until we get to that `else-if` block which will short-circuit on the first successful roll.

Unless you're referring to the inlined code block which was a suggestion for the future implementation

kamronbatman:
So going back through the code it looks like it was done like this on purpose so you only gain 1 stat per skill check. Not sure if that is how OSI does it, but I imagine we need to be careful that we don't gain too many stats at once and go over a cap?

Let's discuss this more in Discord and try to find out the proper logic since I am not 💯 sure.

kamronbatman:
Implemented.

PR #1489, opened 2023-09-03
feat: Adds configurable skill and stat gain

thecatontheflat:
This addresses the following issue: https://github.com/modernuo/ModernUO/issues/1486

Changes:

- Added the following configuration parameters:
   - `skills.gainValueMultiplier` - multiplies the value of the gained Skill
   - `stats.gainValueMultiplier` - multiplies the value of the gained Stat
   - `stats.gainChanceMultiplier` - multiplies the chance of a Stat gain to proc
- Changes to the logic:
   - Added checks to make sure that gain/loss does not exceed the caps nor goes below 0
   - When a Skill has a chance of increasing multiple Stats, all of them will make a roll now (previously it would stop at a first successful roll)
   - Made conditions short-circuit on a simpler condition first before triggering the random number generator

image


Projects/UOContent/Skills/SkillCheck.cs
```diff
@@ -299,17 +347,17 @@ public static void IncreaseStat(Mobile from, Stat stat, bool atrophy)
                     {
                         if (CanLower(from, Stat.Dex) && (from.RawDex < from.RawInt || !CanLower(from, Stat.Int)))
                         {
-                            --from.RawDex;
+                            from.RawDex = DecreaseStatOrZero(from.RawDex);
```
kamronbatman, line 350: The stat property itself already has these checks. So we don't need to do them here.

thecatontheflat:
@kamronbatman It would be best if you take it over, I won't be able to contribute in the next few weeks

kamronbatman:
@thecatontheflat
- Added the 75/25 Primary/Secondary stat for Publish 45 with its stat gain mechanics.
- Removed the stat/skill value gain multipliers since they were ruining the ability to fine-tune skills and decrease/increase accordingly.
- Simplified the logic to prevent going over the gain cap.
- Reduced the stat delay from 15 minutes to 10 minutes per OSI.

Note: I did not add the 15 stats per day limit cause nobody wants a slow game these days and it was removed in UOML.

thecatontheflat:
Thank you!

PR #1492, opened 2023-09-04
Enable multiple characters from the same account

thecatontheflat:


kamronbatman:
I am not sure if that is enough to make this work. Sagas added a similar feature and they had a problem with CUO where it somehow got MUO to log in two NetStates to the same player. So we need specific checks for that.

kamronbatman:
> image
> This is the error CUO received on the logged in character when trying to login into the same character with the current implementation. It logs out the old character and logs in the new one

Yeah that is the happy path, but when CUO is modified to allow multiple characters, it somehow showed the player twice with a small delay between.

kamronbatman:
I did some testing on this. Looks like the CUO bug is an old one.

kamronbatman:
As pointed out in Discord, some systems may enumerate an accounts Mobiles, checking for a NetState to find out if the user is online. This would "break."

So let's leave this as a modification for the developer to do.

PR #1494, opened 2023-09-04
fix: An empty array of tiles is not the same as no water tiles being present

mdodkins:
Boats were completely broken. My server and client are based on 5.0.9.1.

Tracked it down to an empty list erroneously invalidating the sea water check.


kamronbatman:
I think I understand the bug but I am not sure. Can you give an example of how it's broken and where on the map?

PR #1509, opened 2023-09-22
feat: Replaces fastwalk with movement queue for players

kamronbatman:
### Summary

Replaces the leaky bucket fastwalk detection with a queued movement system. Clients will have fast movement queued (up to 100 movement requests) where each movement ack and movement itself is delayed by the movement speed. If a player exceeds the 100 queued movements, then they are disconnected.


kamronbatman:
This won't work since it runs packets out of order.

PR #1530, opened 2023-10-03
fix: Fixes serialization threading, moves world save to end of loop, and eliminates Parallel.ForEach.

kamronbatman:
### Summary
- [X] Eliminates Parallel.ForEach to speed up serializing to memory.
- [X] Adds GenericPersistence as single entities so they are properly serialized in parallel with the new logic.
- [X] Moves saves to the end of the world tick so that way packets/timers/context have all executed.

### DIFF
--- a/Projects/Server/Guild.cs
+++ b/Projects/Server/Guild.cs
@@ -50,9 +50,9 @@ public abstract class BaseGuild : ISerializable
     [CommandProperty(AccessLevel.GameMaster, readOnly: true)]
     public DateTime Created { get; set; } = Core.Now;
 
-    long ISerializable.SavePosition { get; set; } = -1;
+    public long SavePosition { get; set; } = -1;
 
-    BufferWriter ISerializable.SaveBuffer { get; set; }
+    public BufferWriter SaveBuffer { get; set; }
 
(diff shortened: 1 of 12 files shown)


Projects/UOContent/Accounting/Account.cs
```diff
@@ -1205,5 +1205,9 @@ public Mobile Current
                 get => _current;
             }
         }
+
+        public long SavePosition { get; set; }
+
+        public BufferWriter SaveBuffer { get; set; }
```
kamronbatman, line 1211: I am not sure how this worked before without it.

Projects/UOContent/Engines/Bulk Orders/Books/BaseBOBEntry.cs
```diff
@@ -57,4 +57,8 @@ private void Deserialize(IGenericReader reader, int version)
             throw new Exception("Unknown deserialization error in presource generated BOB Entries");
         }
     }
+
+    public long SavePosition { get; set; }
+
+    public BufferWriter SaveBuffer { get; set; }
```
kamronbatman, line 63: Same as above.

PR #1534, opened 2023-10-06
Fix for vendors not actually selling anything (except via Buy Agent)

blendax78:
Issue:
* The Items queue was never actually getting populated with any items, so the `ConsumeTotal` call was always returning false, causing the vendor to always say the user doesn't have enough gold.

Changes:
* Removed the `Item.add()` call and added an `items.Enqueue()` call to items.


kamronbatman (approved):
Thank you!!

PR #1539, opened 2023-10-11
fix: Fixes span reader ReadString

kamronbatman:


blendax78 (approved):
Tested and works great

blendax78 (approved):

PR #1547, opened 2023-10-15
fix: Changes Map.Sector.Items to link list.

kamronbatman:
### Summary

Eliminates `IPooledEnumerable<T>` and `eable.Free()` from `Map` for items. This drastically simplifies code that iterates in range, for example:

```cs
foreach (var item in m.GetItemsInRange(5))
{
}
```
The code above no longer requires an eable and calling `Free()`.

### DIFF
--- a/Projects/Server/Maps/Map.cs
+++ b/Projects/Server/Maps/Map.cs
@@ -38,7 +38,7 @@ public enum MapRules
     FeluccaRules = None
 }
 
-public sealed class Map : IComparable<Map>, ISpanFormattable, ISpanParsable<Map>
+public sealed partial class Map : IComparable<Map>, ISpanFormattable, ISpanParsable<Map>
 {
     public const int SectorSize = 16;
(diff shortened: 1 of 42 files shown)


Projects/Server/Maps/Map.cs
```diff
@@ -1775,4 +1543,194 @@ public static bool TryParse(ReadOnlySpan<char> s, IFormatProvider provider, out
...
+    public class Sector
+    {
+        // TODO: Can we avoid this?
+        private static readonly List<Mobile> m_DefaultMobileList = new();
+        private static readonly List<NetState> m_DefaultClientList = new();
+        private static readonly List<BaseMulti> m_DefaultMultiList = new();
+        private static readonly List<Region> m_DefaultRectList = new();
+        private bool m_Active;
+        private List<NetState> _clients;
+        private ValueLinkList<Item> _items;
```
kamronbatman, line 1564: Crux of the change is here.

PR #1553, opened 2023-10-17
fix: Changes Map.Sector.Mobiles to link list & Fixes various related crash bugs

kamronbatman:
### Summary

Eliminates `IPooledEnumerable<T>` and `eable.Free()` from `Map` for mobiles. This drastically simplifies code that iterates in range, for example:

```cs
foreach (var m in m.GetMobilesInRange(5))
{
}
```
The code above no longer requires an eable and calling `Free()`.

- [X] Fixed several locations where an NPC that was damaged would cause a server crash.
- [X] Removed an unnecessary allocation in guard fake calls (NPCs calling guards on you)
- [X] Fixes damage precision loss in Poison Strike Spell
- [X] BogThing no longer attempts to "search" for boglings to eat when it is at full health

### DIFF
--- a/Projects/UOContent/Regions/GuardedRegion.cs
+++ b/Projects/UOContent/Regions/GuardedRegion.cs
@@ -153,8 +153,14 @@ public class GuardedRegion : BaseRegion
 
     public override void MakeGuard(Mobile focus)
     {
-        var eable = focus.GetMobilesInRange<BaseGuard>(8);
-        var useGuard = eable.FirstOrDefault(m => m.Focus == null);
+        BaseGuard useGuard = null;
+        foreach (var m in focus.GetMobilesInRange<BaseGuard>(8))
+        {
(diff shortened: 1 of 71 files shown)


Projects/UOContent/Regions/GuardedRegion.cs
```diff
@@ -286,19 +292,8 @@ public void CheckGuardCandidate(Mobile m)
...
-                        1013037,
-                        1013038,
-                        1013039,
-                        1013041,
-                        1013042,
-                        1013043,
-                        1013052
-                    )
-                );
+                fakeCall.Say(Utility.Random(1013037, 16));
```
kamronbatman, line 295: - Removed `1007037` because it is from an older system of NPCs that you fight that would spam you with stupid messages that nobody cared about.
- Removed `501603` since it is a copy of another one we already have.
- Added some missing messages.

PR #1563, opened 2023-10-23
fix: FlippableAttribute with empty list will crash the client.

blendax78:
Using a decorator tool on certain items (example: some Paragon Chests) will cause clients to crash, throwing a `IndexOutOfRangeException`.

It depends on the ItemIDs property, and we were only checking if it's null or not, not if it has any elements, which ended up causing it to set `item.ItemID = ItemIDs[index];` (or null).


kamronbatman:
Great catch. I think this never came up since `[Flippable]` with no arguments shouldn't normally happen. Do we have that in our code?

Also, I am thinking we should check for an empty array in the constructor of the attribute and throw. Thoughts?

kamronbatman (approved):

PR #1585, opened 2023-11-09
Fix for missing ethereal horse label

blendax78:
image

Fixed missing label for ethereal horse statues.  Issue can be seen in screen shot.


kamronbatman:
I probably deleted it on accident when I code genned it. Thank you!

PR #1605, opened 2023-11-20
Fix Server.TryParse so it works for Boolean types

mdodkins:
Server.TryParse() function was failing to parse "True" or "False" as bools. I've fixed this and added tests.

### DIFF
--- a/Projects/UOContent.Tests/Tests/Utilities/TryParseTests.cs
+++ b/Projects/UOContent.Tests/Tests/Utilities/TryParseTests.cs
@@ -1,0 +1,21 @@
+using Server;
+using System;
+using Xunit;
+
+namespace Server.Tests.Utility;
+
+public class TryParseTests
+{
+    [Theory]
+    [InlineData("True", null, true)]
+    [InlineData("False", null, false)]
+    [InlineData("Alakazam", "Bool parse failed", true)]
+    public void TestTryParseBool(string value, string returned, bool parsedAs)
+    {
+        string actualReturned = Server.Types.TryParse(typeof(bool), value, out object constructed);
+        Assert.Equal(returned, actualReturned);
+
+        if (returned == null)
+            Assert.Equal(parsedAs, constructed);
+    }
+}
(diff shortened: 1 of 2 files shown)


Projects/UOContent.Tests/Tests/Utilities/TryParseTests.cs
```diff
@@ -0,0 +1,21 @@
...
+    [InlineData("True", null, true)]
+    [InlineData("False", null, false)]
+    [InlineData("Alakazam", "Not a valid boolean string.", true)]
+    public void TestTryParseBool(string value, string returned, bool parsedAs)
+    {
+        string actualReturned = Server.Types.TryParse(typeof(bool), value, out object constructed);
+        Assert.Equal(returned, actualReturned);
+
+        if (returned == null)
+            Assert.Equal(parsedAs, constructed);
```
kamronbatman, line 19:
```suggestion
        if (returned == null)
        {
            Assert.Equal(parsedAs, constructed);
        }
```

Projects/UOContent/Utilities/Types.cs
```diff
@@ -187,6 +187,18 @@ public static string TryParse(Type type, string value, out object constructed)
                 return null;
             }
 
+            if (IsType(type, OfBool))
+            {
+                bool parsed;
+                bool parseSuccessful = bool.TryParse(value, out parsed);
```
kamronbatman, line 193:
```suggestion
                bool parseSuccessful = bool.TryParse(value, out bool parsed);
```

Projects/UOContent/Utilities/Types.cs
```diff
@@ -187,6 +187,18 @@ public static string TryParse(Type type, string value, out object constructed)
...
+                if (!parseSuccessful)
+                {
+                    return "Not a valid boolean string.";
+                }
+                constructed = parsed;
+                return null;
```
kamronbatman, line 199:
```suggestion
                if (bool.TryParse(value, out bool parsed))
                {
                    constructed = parsed;
                    return null;
                }
                
                return "Not a valid boolean string.";
```

kamronbatman (approved):

kamronbatman:
Thank you!!

PR #1610, opened 2023-11-23
fix: Fixes TimeSpan rounding issues

### LINKED ISSUE #1604: Potential issue with TimeSpan factory methods
It appears that TimeSpan factory methods auto-rounded their results in the .net framework. Once moving to .NET 5+ we noticed issues with buff icons not expiring correctly because now TimeSpan factory methods do no such thing. 

https://trueuo.com/threads/new-sort-of-lags.3699/page-2#post-20745

This is how I resolved it for now, regarding buff icons, but there could be other underlying systems affected since MUO was written off of RunUO which may have used TimeSpan factory methods under the assumption they rounded.
https://github.com/TrueUO/TrueUO/pull/1280

kamronbatman:
### Summary

TLDR - .NET Framework rounded TimeSpan (and some DateTime) methods to the nearest millisecond. This was actually _expected_ accidentally for parts of RunUO.

We are aiming to fix this and clean up some other bad assumptions.

Closes #1604

### DIFF
--- a/Projects/UOContent/Items/Skill Items/Misc/Bandage.cs
+++ b/Projects/UOContent/Items/Skill Items/Misc/Bandage.cs
@@ -495,35 +495,32 @@ public class BandageContext : Timer
                     seconds = 9.4 + 0.6 * ((double)(120 - dex) / 10);
                 }
             }
-            else
+            else if (Core.AOS && GetPrimarySkill(patient) == SkillName.Veterinary)
             {
-                if (Core.AOS && GetPrimarySkill(patient) == SkillName.Veterinary)
+                seconds = 2.0;
+            }
+            else if (Core.AOS)
+            {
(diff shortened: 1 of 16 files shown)


Projects/UOContent/Items/Skill Items/Misc/Bandage.cs
```diff
@@ -495,36 +495,33 @@ public static BandageContext BeginHeal(Mobile healer, Mobile patient)
                     seconds = 9.4 + 0.6 * ((double)(120 - dex) / 10);
                 }
             }
-            else
+            else if (Core.AOS && GetPrimarySkill(patient) == SkillName.Veterinary)
```
kamronbatman, line 498: No functional change in this file, just collapsed the else and if into an else-if.

PR #1615, opened 2023-11-25
fix: Fixes character starting cities.

### LINKED ISSUE #1408: limit new haven to pub 16
Creating a character on a pre Renaissance server take it to haven.

kamronbatman:
### Summary

- Moves starting city info from AccountHandler to CharacterCreation
- Simplifies the logic of determining the starting cities
- Adds support for Trammel & Felucca for non-young accounts
- Fixes fall through starting city for v6+ in UOR era.
- All staff are force-sent to GA.

### TODOS

- Move the starting cities to JSON files
- Add in New Magincia (`1075077`) to the client versions/eras that support it.

Closes #1408

### DIFF
--- a/Projects/UOContent/Misc/CharacterCreation.cs
+++ b/Projects/UOContent/Misc/CharacterCreation.cs
@@ -181,6 +181,13 @@ public static class CharacterCreation
             Array.Copy(FeluccaStartingCities, 0, cities, index, FeluccaStartingCities.Length);
         }
 
+        // TODO: Add Royal City for gargoyles
+
+        if (cities.Length == 0)
+        {
+            logger.Warning("Both Felucca and Trammel are unavailable maps, therefore no starting cities are available.");
+        }
+
         return cities;
     }
 
--- a/Projects/UOContent/Misc/CharacterCreation.cs
+++ b/Projects/UOContent/Misc/CharacterCreation.cs
@@ -136,19 +136,9 @@ public static class CharacterCreation
 
     public static CityInfo[] GetStartingCities(bool isYoung)
     {
-        if (isYoung)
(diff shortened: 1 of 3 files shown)


Projects/UOContent/Misc/CharacterCreation.cs
```diff
@@ -76,8 +76,111 @@ public static class CharacterCreation
...
+        }
+
+        if (feluccaAvailable)
+        {
+            Array.Copy(FeluccaStartingCities, 0, cities, index, FeluccaStartingCities.Length);
+        }
+
+        // TODO: Add Royal City for gargoyles
+
+        if (cities.Length == 0)
```
blendax78, line 177: Should this be moved up to line 161, and maybe have a return statement if `cities.length == 0`?  Just wondering if there are no cities available, do we need to do the Array.Copy?  Also, will it cause issues if (somehow) TrammelStartingCities.length? == 0 AND FeluccaStartingCities.Length == 0, and we try to do the Array.Copy??

kamronbatman: Yeah good call! - And yeah, it will cause issues, hence the warning. Specifically, players won't be able to log in. We don't want to send them to a map that isn't available, for whatever reason. So we either have to crash, or throw a warning/error.

Thoughts on other options?

blendax78: ahh ok.  Yeah, I can't really think of any other good way to handle that.  Like you said, throwing a warning/error is probably the best bet

blendax78 (commented):
💯 awesome

mdodkins:
Looks good :+1:

PR #1635, opened 2023-12-14
fix: Client crash when banker tries to use checks & gold for a purchase

blendax78:
Discovered when players would use a House Placement Tool to make a large purchase. When this code would check their bank box:
1. It wouldn't recursively check containers in the bank box for gold/checks.
2. It would crash when players had check(s) and gold in their bank box. Full Disclaimer: I'm by no means a .NET expert. I don't know how the code works that I removed (Feel free to explain!). It seemed to act like the two variables were references to one another. When my test character had 2 piles of gold and 2 checks in their main bank box, it would return both `gold` AND `checks` as lists with FOUR objects each. This caused a client crash in Banker.cs line 138, when it was attempting to cast a `Gold` object into a `BankCheck` object. It most likely caused issues on line 122, as well, since the `gold` variable was being looped through (I didn't test that, but it's a pretty good guess 😄 ).

### DIFF
--- a/Projects/UOContent/Mobiles/Townfolk/Banker.cs
+++ b/Projects/UOContent/Mobiles/Townfolk/Banker.cs
@@ -66,7 +66,8 @@ public partial class Banker : BaseVendor
     public static int GetBalance(Mobile m, out List<Item> gold, out List<Item> checks)
     {
         long balance = 0;
-        gold = checks = new List<Item>();
+        var gold = new List<Item>();
+        var checks = new List<Item>();
 
         if (AccountGold.Enabled && m.Account != null)
         {
@@ -82,7 +83,7 @@ public partial class Banker : BaseVendor
 
         if (bank != null)
         {
-            foreach (var g in bank.FindItemsByType<Gold>())
+            foreach (var g in bank.FindItemsByType<Gold>(true))
             {
                 balance += g.Amount;
                 gold.Add(g);
@@ -93,7 +94,7 @@ public partial class Banker : BaseVendor
                 return int.MaxValue;
             }
 
-            foreach (var bc in bank.FindItemsByType<BankCheck>())
+            foreach (var bc in bank.FindItemsByType<BankCheck>(true))
             {
                 balance += bc.Worth;
                 checks.Add(bc);
--- a/Projects/UOContent/Mobiles/Townfolk/Banker.cs
+++ b/Projects/UOContent/Mobiles/Townfolk/Banker.cs
@@ -126,7 +126,7 @@ public partial class Banker : BaseVendor
         gold = PooledRefList<Gold>.Create();
         checks = PooledRefList<BankCheck>.Create();
 
-        foreach (var g in bank.FindItemsByType<Gold>(true))
+        foreach (var g in bank.FindItemsByType<Gold>())
         {
             balance += g.Amount;
             gold.Add(g);
@@ -137,7 +137,7 @@ public partial class Banker : BaseVendor
             }
         }
 
-        foreach (var bc in bank.FindItemsByType<BankCheck>(true))
+        foreach (var bc in bank.FindItemsByType<BankCheck>())
         {
             balance += bc.Worth;
             checks.Add(bc);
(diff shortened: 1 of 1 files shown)


blendax78:
Whomp whomp. I'm still on SDK 7.0.0, so I can't build the current main.

kamronbatman:
Thanks! I'll take a look and clean it up. Meanwhile, get .net 8 sdk!

Projects/UOContent/Mobiles/Townfolk/Banker.cs
```diff
@@ -82,7 +84,9 @@ public static int GetBalance(Mobile m, out List<Item> gold, out List<Item> check
 
         if (bank != null)
         {
-            foreach (var g in bank.FindItemsByType<Gold>())
+            gold = new List<Gold>();
+
+            foreach (var g in bank.FindItemsByType<Gold>(true))
```
kamronbatman, line 89: The argument is true by default, so this shouldn't "fix" the recurse issue?

Projects/UOContent/Mobiles/Townfolk/Banker.cs
```diff
@@ -103,6 +109,51 @@ public static int GetBalance(Mobile m, out List<Item> gold, out List<Item> check
...
+            gold = default;
+            checks = default;
+
+            return false;
+        }
+
+        long balance = 0;
+
+        gold = PooledRefList<Gold>.Create();
+        checks = PooledRefList<BankCheck>.Create();
```
kamronbatman, line 127: Let's move this to LIne 140 (above the bankcheck foreach).
Then above line 136 (the return true), let's add `checks = default;`

Projects/UOContent/Mobiles/Townfolk/Banker.cs
```diff
@@ -103,6 +109,51 @@ public static int GetBalance(Mobile m, out List<Item> gold, out List<Item> check
         return (int)Math.Clamp(balance, 0, int.MaxValue);
     }
 
+    private static bool HasRequiredBalance(int requiredBalance, Mobile m, out PooledRefList<Gold> gold, out PooledRefList<BankCheck> checks)
```
kamronbatman, line 112: We can probably get away with a `PooledRefQueue` - Which I think is slightly more performant? - I would keep it a list if this function was public. 🤔

Projects/UOContent/Mobiles/Townfolk/Banker.cs
stefanomerotta: Removed the explicit flag since there are no differences. 
From latest test the recursion is working

blendax78:
Screenshot 2023-12-14 at 9 35 56 PM


Screenshot 2023-12-14 at 9 35 37 PM

Projects/UOContent/Mobiles/Townfolk/Banker.cs
```diff
@@ -103,6 +109,51 @@ public static int GetBalance(Mobile m, out List<Item> gold, out List<Item> check
...
+            balance += g.Amount;
+            gold.Add(g);
+
+            if (balance >= requiredBalance)
+            {
+                return true;
+            }
+        }
+
+        foreach (var bc in bank.FindItemsByType<BankCheck>())
```
kamronbatman, line 140:
```suggestion
        foreach (var g in bank.FindItemsByType<Gold>())
        {
            balance += g.Amount;
            gold.Add(g);

            if (balance >= requiredBalance)
            {
                checks = default;
                return true;
            }
        }

        checks = PooledRefList<BankCheck>.Create();
 
        foreach (var bc in bank.FindItemsByType<BankCheck>())
```

Projects/UOContent/Mobiles/Townfolk/Banker.cs
kamronbatman, line 89:
```suggestion
            foreach (var g in bank.FindItemsByType<Gold>())
```

Projects/UOContent/Mobiles/Townfolk/Banker.cs
```diff
@@ -93,7 +97,9 @@ public static int GetBalance(Mobile m, out List<Item> gold, out List<Item> check
                 return int.MaxValue;
             }
 
-            foreach (var bc in bank.FindItemsByType<BankCheck>())
+            checks = new List<BankCheck>();
+
+            foreach (var bc in bank.FindItemsByType<BankCheck>(true))
```
kamronbatman, line 102:
```suggestion
            foreach (var bc in bank.FindItemsByType<BankCheck>())
```

PR #1645, opened 2023-12-22
fix: Use pooled ref for corpse instanced loot

kamronbatman:


Projects/UOContent/Items/Misc/Corpses/Corpse.cs
```diff
@@ -382,29 +383,23 @@ private void AssignInstancedLoot()
...
 
-        var attackers = new List<Mobile>(_aggressors);
-
-        for (var i = 1; i < attackers.Count - 1; i++) // randomize
-        {
-            var rand = Utility.Random(i + 1);
-
-            (attackers[rand], attackers[i]) = (attackers[i], attackers[rand]);
-        }
+        using var attackers = PooledRefList<Mobile>.Create();
```
kamronbatman, line 395: Use _aggressors.Count as a hint to the minimum size.

PR #1649, opened 2023-12-28
feat: Adds AdvancedSearch (XmlFind)

kamronbatman:
### Summary

Adds Advanced Search (XmlFind) using `[AS` command.

Notable differences from XmlFind:
* No save to file.
* No "Display from" option (might add that later).
* No advanced XmlSpawner property test syntax. Only basic property operator comparisons.
* This version will freeze your shard, but fans out to all processors just like world saves. (So it should be stupid fast, especially if you have lots of cores)
* Does not support spawn entry/type/error searching. We have a spawn search command to do that already.
* I haven't set up correct "defaults" for the options when you first open it. I'll do that later.

### Screenshots
image

image

image

### DIFF
--- a/Projects/UOContent/Engines/Advanced Search/AdvancedSearchGump.cs
+++ b/Projects/UOContent/Engines/Advanced Search/AdvancedSearchGump.cs
@@ -1,0 +1,962 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using CommunityToolkit.HighPerformance;
+using Server.Commands;
+using Server.Commands.Generic;
+using Server.Engines.Spawners;
+using Server.Gumps;
+using Server.Network;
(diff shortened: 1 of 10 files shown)


Projects/UOContent/Engines/Advanced Search/AdvancedSearchGump.cs
```diff
@@ -0,0 +1,888 @@
...
+                    comparer = SortDescending
+                        ? AdvancedSearchResultSelectedComparer.InstanceReverse
+                        : AdvancedSearchResultSelectedComparer.Instance;
+                    break;
+                }
+        }
+
+        if (SearchResults?.Length > 0)
+        {
+            Array.Sort(SearchResults, comparer);
```
kamronbatman, line 791: Technically we could sort in a background thread. Would we ever need to? Probably not?

PR #1652, opened 2023-12-30
feat: Adds optimized dynamic/static layout gumps

### LINKED ISSUE #1640: Gump refactor for caching and functional generation
This issue aims to track the conversion of gump generation to reduce object and string allocations and allow caching for gumps with static content.

Ongoing development is on dedicated branch:

kamronbatman:
> [!Important]
> In the future, gumps that come with ModernUO will be converted to this new API where it makes sense. Please reach out in [discord](https://muo.gg/discord) if you have questions or need help!

> [!Note]
> **DEVELOPER NOTE**
> We will not deprecate the legacy (RunUO) gump API. If you want to continue to use it, go for it!

# New Gump API
We are pleased to release a new API that is faster, allocates nearly zero memory, and still feels very similar to the original API. The API is broken into 3 types of gumps, dynamic, static with placeholders, and static without placeholders. 

### Dynamic Gumps
These gumps will inherit `DynamicGump` and are meant for gumps that have a dynamic layout. This includes specifying dynamic arguments to HtmlLocalized entries.

<details>
  <summary>Example</summary>

  ```cs
public class PetResurrectGump : DynamicGump
{
    private readonly string _petName;

    public PetResurrectGump(string petName) : base(50, 50)
    {
        from.CloseGump<PetResurrectGump>();

        _petName = petName;
    }

    protected override void BuildLayout(ref DynamicGumpBuilder builder)
    {
        builder.AddPage();

        builder.AddBackground(10, 10, 265, 140, 0x242C);

        builder.AddItem(205, 40, 0x4);
        builder.AddItem(227, 40, 0x5);

        builder.AddItem(180, 78, 0xCAE);
        builder.AddItem(195, 90, 0xCAD);
        builder.AddItem(218, 95, 0xCB0);

        // <div align=center>Wilt thou sanctify the resurrection of:</div>
        builder.AddHtmlLocalized(30, 30, 150, 75, 1049665);
        builder.AddHtml(30, 70, 150, 25, _petName, true);

        builder.AddButton(40, 105, 0x81A, 0x81B, 0x1);  // Okay
        builder.AddButton(110, 105, 0x819, 0x818, 0x2); // Cancel
    }

    public override void OnResponse(NetState state, in RelayInfo info)
    {
        // Normal on response code that all gumps have.
    }
}
  ```
</details>

## Static Gumps 
Static gumps are those where the function to the build the layout is called only once and cached forever. They can optionally have placeholders. These placeholders allow the developer to specify the string values later, dynamically in a `BuildStrings` method on the gump. If a gump does not have any placeholders, the string entries will also be cached forever.

<details>
  <summary>Example</summary>

  ```cs
public class PetResurrectGump : StaticGump<PetResurrectGump>
{
    private readonly string _petName;

    public PetResurrectGump(string petName) : base(50, 50)
    {
        from.CloseGump<PetResurrectGump>();

        _petName = petName;
    }

    protected override void BuildLayout(ref StaticGumpBuilder builder)
    {
        builder.AddPage();

        builder.AddBackground(10, 10, 265, 140, 0x242C);

        builder.AddItem(205, 40, 0x4);
        builder.AddItem(227, 40, 0x5);

        builder.AddItem(180, 78, 0xCAE);
        builder.AddItem(195, 90, 0xCAD);
        builder.AddItem(218, 95, 0xCB0);

        // <div align=center>Wilt thou sanctify the resurrection of:</div>
        builder.AddHtmlLocalized(30, 30, 150, 75, 1049665);
        builder.AddHtmlPlaceholder(30, 70, 150, 25, "petName", true);

        builder.AddButton(40, 105, 0x81A, 0x81B, 0x1);  // Okay
        builder.AddButton(110, 105, 0x819, 0x818, 0x2); // Cancel
    }

    protected override void BuildStrings(ref GumpStringsBuilder builder)
    {
        builder.SetStringSlot("petName", $"<CENTER>{_pet.Name}</CENTER>");
    }

    public override void OnResponse(NetState state, in RelayInfo info)
    {
        // Normal on response code that all gumps have.
    }
}
  ```
</details>

## Benchmarks
To make sure we were going in the right direction and not wasting time, we took copious benchmarks. Here are the final benchmarks for a really simple gump. 

Note:
* The majority of creating a gump is compressing the layout and the strings. Compressing each section takes ~6,000ns (12us total).

```cs
| Method                         | Mean         | Error        | StdDev       | Median       | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------------------- |-------------:|-------------:|-------------:|-------------:|------:|--------:|-------:|----------:|------------:|
| OldGump                        | 13,308.29 ns | 1,059.695 ns | 1,883.608 ns | 14,330.72 ns | 1.000 |    0.00 | 0.1526 |    2400 B |        1.00 |
| DynamicLayoutGump              | 13,357.86 ns |   129.144 ns |   226.185 ns | 13,323.60 ns | 1.029 |    0.17 |      - |      48 B |        0.02 |
| StaticLayoutDynamicStringsGump |  6,653.10 ns |    81.815 ns |   143.292 ns |  6,617.45 ns | 0.514 |    0.09 |      - |      40 B |        0.02 |
| StaticLayoutGump               |     86.33 ns |     0.760 ns |     1.350 ns |     86.07 ns | 0.007 |    0.00 | 0.0020 |      32 B |        0.01 |
```

# Non-Breaking Changes
* All gump components in the core have been moved to `Gumps/Legacy`.
* All legacy gumps will still inherit `Gump`, which now inherits `BaseGump`

# Special Thanks

Thank you to @stefanomerotta for considerable contributions/benchmarking/testing to make this effort a reality! We collectively went through over 10 iterations, but it is finally ready.

Closes #1640


Projects/Server/Gumps/New/Components/Interpolation/GumpInterpolatedStringHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+using static Server.Gumps.Components.Interpolation.GumpInterpolatedStringHandler;
+
+namespace Server.Gumps.Components.Interpolation
+{
+    [InterpolatedStringHandler]
+    public ref struct GumpInterpolatedStringHandler<TStringHandler, TFormatter>
+        where TStringHandler : struct, IStringsHandler
+        where TFormatter : struct, IGumpInterpolationTextFormatter<TFormatter>
+    {
+        private static readonly char[] _buffer = GC.AllocateUninitializedArray<char>(1024);
```
kamronbatman, line 29: I think this gets replicated for reach concrete type. At least that is how it works for classes. We have to find out if it works the same with ref structs (since that is a new construct), or if it is done by erasure.

Projects/Server/Gumps/New/DynamicGridGump.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                type = 0;
+                index = 0;
+                return false;
+            }
+
+            buttonID -= 1;
+
+            type = buttonID % typeCount;
+            index = buttonID / typeCount;
```
kamronbatman, line 71: nit:

```suggestion
            index = Math.DivRem(buttonID, typeCount, out var type);
```

Projects/Server/Gumps/New/StaticGump.cs
```diff
@@ -0,0 +1,82 @@
...
+        protected abstract void GetGumpData(out LayoutEntry layout, out StringsEntry strings);
+
+        private void Send(NetState ns, in LayoutEntry layout, in StringsEntry strings)
+        {
+            int worstLayoutLength = Zlib.MaxPackSize(layout.UncompressedLength);
+            int worstStringsLength = Zlib.MaxPackSize(strings.UncompressedLength);
+
+            int maxLength = 40 + worstLayoutLength + worstStringsLength;
+
+            SpanWriter writer = new(maxLength);
```
kamronbatman, line 46: Can we use a using here?
```suggestion
            using var writer = new SpanWriter(maxLength);
```

Then we can get rid of the dispose on line 79 too.

Projects/UOContent/Network/Packets/IncomingPlayerPackets.cs
```diff
@@ -420,27 +424,36 @@ public static void DisplayGumpResponse(NetState state, SpanReader reader, int pa
...
-                    // state.Disconnect("Invalid gump response.");
-                    return;
-                }
+                    if (textLength > 239)
+                    {
+                        state.LogInfo("Invalid gump response, disconnecting...");
+                        var exception = new InvalidGumpResponseException($"Text entry {i} is too long ({textLength})");
+                        exception.SetStackTrace(new StackTrace());
+                        NetState.TraceException(exception);
+                        state.Mobile?.SendMessage("Invalid gump response.");
```
kamronbatman, line 448: We used to do this, but we don't want to make it easier for players to poke holes in badly created gumps, so we opt to disconnect them.

Projects/Server/Gumps/BaseGump.cs
```diff
@@ -0,0 +1,64 @@
...
+            TypeID = GetTypeId(GetType());
+        }
+
+        public abstract void SendTo(NetState ns);
+
+        public virtual void OnResponse(NetState sender, in RelayInfo info)
+        { }
+
+        public virtual void OnServerClose(NetState owner)
+        { }
```
kamronbatman, line 56:
```suggestion
        public virtual void OnResponse(NetState sender, in RelayInfo info)
        {
        }

        public virtual void OnServerClose(NetState owner)
        {
        }
```

Projects/Server/Gumps/New/StaticGump.cs
```diff
@@ -0,0 +1,82 @@
...
+            writer.Write((byte)0xDD); // Packet ID
+            writer.Seek(2, SeekOrigin.Current);
+
+            writer.Write(Serial);
+            writer.Write(TypeID);
+            writer.Write(X);
+            writer.Write(Y);
+
+            if (!layout.Compressed)
```
kamronbatman, line 55: Should be flipped? If it is compressed, then write packed?

Projects/Server/Gumps/New/StaticGump.cs
```diff
@@ -0,0 +1,82 @@
...
+                OutgoingGumpPackets.WritePacked(layout.Data, ref writer);
+            }
+            else
+            {
+                writer.Write(layout.Data);
+            }
+
+            writer.Write(strings.Count);
+
+            if (!strings.Compressed)
```
kamronbatman, line 66: Should be flipped? If it is compressed, then write packed?

stefanomerotta: No, it means: if the given strings byte array has already been compressed you can simply write it as is into the resulting packet, otherwise you have to call "WritePacked" that will compress them and then write the compressed buffer into the resulting packet.

Projects/Server/Gumps/New/DynamicGridGump.cs
stefanomerotta: Done!

Projects/Server/Gumps/New/StaticGump.cs
stefanomerotta: Same as below comment

Projects/Server/Gumps/New/Components/Interpolation/GumpInterpolatedStringHandler.cs
stefanomerotta: Moved outside the generic class into file scoped static class

stefanomerotta:
@kamronbatman I've just refactor the StaticGump as discussed yesterday.
I also removed the DynamicStringPlaceholder and added a bunch of extension methods for dynamic string placeholder (they have the suffix "Slot" in the name)

The converted static gump with dynamic strings is "PetResurrectGump"

Projects/Server/Gumps/New/Components/GridGumpBuilder.cs
```diff
@@ -0,0 +1,284 @@
...
+        private int _currentY;
+        private int _backgroundWidth;
+        private int _offsetWidth;
+        private ushort _currentPage;
+
+        public readonly int CurrentX => _currentX;
+        public readonly int CurrentY => _currentY;
+        public readonly ReadOnlySpan<byte> Layout => _builder.Layout;
+        public readonly int LayoutSize => _builder.LayoutSize;
+        [UnscopedRef] public ref T StringsWriter => ref _builder.StringsWriter;
```
kamronbatman, line 56:
```suggestion
        [UnscopedRef]
        public ref T StringsWriter => ref _builder.StringsWriter;
```

Projects/Server/Gumps/BaseGump.cs
```diff
@@ -0,0 +1,66 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>. *
+ *************************************************************************/
+
+using Server.Network;
+using System;
+using System.Runtime.CompilerServices;
+
+namespace Server.Gumps
+{
+    public abstract class BaseGump
```
kamronbatman, line 22: @Voxpire is worried about name clashing. Let's see if we can come up with a better name?

Projects/Server/Gumps/New/Components/GridGumpBuilder.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            if (_backgroundWidth == 0)
+            {
+                return;
+            }
+
+            int backgroundHeight = _currentY + _entryHeight + _offsetSize + _borderSize;
+            int offsetHeight = _currentY + _entryHeight + _offsetSize - _borderSize;
+
+            Span<char> buffer = stackalloc char[46 * 2];
```
kamronbatman, line 86: How did we come to the number 46 *2?
My estimate is that at most (which it never would be), it would be 8 characters per number, so 24 + 20 for the literals.
So 44 characters (no need to times it by 2 cause it is already chars?)

Projects/Server/Gumps/New/Components/GridGumpBuilder.cs
```diff
@@ -0,0 +1,285 @@
...
+
+            MemoryExtensions.TryWrite(buffer, $"{{ resizepic 0 0 {_backGumpId} {_backgroundWidth} {backgroundHeight} }}", out int charsWritten);
+            OperationStatus result = Ascii.FromUtf16(buffer[..charsWritten], _backgroundSpan, out int bytesWritten);
+            _backgroundSpan[bytesWritten..].Fill((byte)' ');
+
+            Debug.Assert(result == OperationStatus.Done);
+
+            MemoryExtensions.TryWrite(buffer, $"{{ gumppictiled {_borderSize} {_borderSize} {_offsetWidth} {offsetHeight} {_offsetGumpId} }}", out charsWritten);
+            result = Ascii.FromUtf16(buffer[..charsWritten], _offsetSpan, out bytesWritten);
+            _offsetSpan[bytesWritten..].Fill((byte)' ');
```
kamronbatman, line 96: Why do we need to fill it with spaces?

Projects/Server/Gumps/New/Components/GridGumpBuilder.cs
kamronbatman, line 86: Looks like the buffer is used two separate times. I don't think we need to multiply it by 2.
The second usage is the biggest, so my estimate is 21 for literals, up to 5 characters per integer, so 40 for integers. That would be 61 total?

Projects/Server/Gumps/New/DynamicGridGump.cs
```diff
@@ -0,0 +1,128 @@
...
+            SpanWriter writer = new(maxLength);
+            writer.Write((byte)0xDD); // Packet ID
+            writer.Seek(2, SeekOrigin.Current);
+
+            writer.Write(Serial);
+            writer.Write(TypeID);
+            writer.Write(_x);
+            writer.Write(_y);
+
+            builder.FinalizeLayout();
```
kamronbatman, line 115: All of the calls are throwing a warning saying the method is impure and executed against a readonly ref, causing a automatic defensive copy.

Projects/Server/Gumps/New/Components/GumpBuilder.cs
```diff
@@ -0,0 +1,698 @@
...
+        }
+
+        public void AddItemProperty(Serial serial)
+        {
+            WriteStart(Labels.ItemProperty);
+            WriteValue(serial.Value);
+            WriteEnd();
+        }
+
+        public void AddLabel(int x, int y, int hue, ReadOnlySpan<char> text)
```
kamronbatman, line 351: For everything that takes text, let's add (in addition to ReadOnlySpan<char>), let's add support for RawInterpolatedStringHandler. The advantage of that is when we do this:

```cs
builder.AddLabel(x, y, hue, $"key_{slotKey}");
```

We are not creating an intermediary string that is allocated and discarded. The `RawInterpolatedStringHandler` uses a rented buffer. See `SpanWriter` for how to use it, it is like the other ones.

Projects/Server/Gumps/New/Components/GumpBuilder.cs
```diff
@@ -0,0 +1,698 @@
...
+            WriteValue(height);
+            WriteValue(number);
+            WriteValue(background);
+            WriteValue(scrollbar);
+            WriteEnd();
+        }
+
+        public void AddHtmlLocalized(int x, int y, int width, int height, int number, short color, bool background = false, bool scrollbar = false)
+        {
+            WriteStart(Labels.HtmlLocalizedWithColor);
```
kamronbatman, line 239: Were you able to find out if doing this:
```suggestion
            WriteStart("xmfhtmlgumpcolor"u8);
```

Creates an allocation, or does it use a generated global constant?

Projects/Server/Gumps/New/Components/GumpBuilder.cs
```diff
@@ -0,0 +1,698 @@
...
+        {
+            DynamicStringsHandler.DynamicMode = true;
+            builder.AddTextEntryLimited(x, y, width, height, hue, entryId, $"key_{slotKey}", size);
+            DynamicStringsHandler.DynamicMode = false;
+        }
+    }
+
+    static file class Properties
+    {
+        public static readonly byte[] Buffer = Encoding.ASCII.GetBytes("{ nomove }{ noclose }{ nodispose }{ noresize }");
```
kamronbatman, line 663: Let's also switch this to u8's like the others.

Projects/Server/Gumps/New/Components/GumpBuilder.cs
```diff
@@ -0,0 +1,698 @@
...
+using System.Text;
+using static Server.Gumps.Components.Interpolation.GumpInterpolatedStringHandler;
+using static Server.Gumps.Components.GumpBuilderExtensions;
+
+namespace Server.Gumps.Components
+{
+    public ref struct GumpBuilder<T>
+        where T : struct, IStringsHandler
+    {
+        private static readonly byte[] _layoutBuffer = LayoutBuffer;
```
kamronbatman, line 35: We need to find out if we can use statics in generic ref structs. We may have to just use `LayoutBuffer` directly or use `STArrayPool`.

Projects/Server/Gumps/New/Components/Interpolation/GumpInterpolatedStringHandler.Formatters.cs
```diff
@@ -0,0 +1,75 @@
...
+ * the Free Software Foundation, either version 3 of the License, or     *
+ * (at your option) any later version.                                   *
+ *                                                                       *
+ * You should have received a copy of the GNU General Public License     *
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>. *
+ *************************************************************************/
+
+namespace Server.Gumps.Components.Interpolation
+{
+    public static class GumpInterpolatedStringHandler
```
kamronbatman, line 18: We have a static class and ref struct with the same name. I am not sure how this even compiles? LOL - Should we distinguish them? Do we even need this top level class?

stefanomerotta: Yep, this compiles because the ref struct is a generic one, so the type doesn't collide.
I renamed the static class and moved out the related interface

Projects/Server/Gumps/New/Components/GumpBuilder.cs
stefanomerotta: I forgot to delete the old buffer, it it not used anymore: I just deleted it

Projects/Server/Gumps/New/Components/GridGumpBuilder.cs
kamronbatman: I understand why now.

Projects/Server/Gumps/New/Components/GumpBuilder.cs
kamronbatman: I figured out the issue. I'll make adjustments.

Projects/Server/Gumps/New/Components/GridGumpBuilder.cs
stefanomerotta: You are right: fixed the wrong buffer size multiplication

Projects/Server/Gumps/New/DynamicGridGump.cs
stefanomerotta: Good catch: I fixed all the wrong readonly ref  parameters for new gumps

Projects/Server/Gumps/New/Components/GumpBuilder.cs
stefanomerotta: That is simply a static "local" variable that refers to the same global buffer.
If you prefer we can reference the global variable directly, if we want to maintain the static global buffer, or use the rented buffer.
The actual logic is similar to the old gump approach with a huge global static buffer.

Projects/Server/Gumps/New/Components/GumpBuilder.cs
stefanomerotta: u8 strings are converted into global static byte array under the hood.
From my tests, the instruction:
```cs
AppendFake("test"u8);
```
Will be converted into these IL:
```
IL_0000: ldsflda valuetype '<PrivateImplementationDetails>'/'__StaticArrayInitTypeSize=5' '<PrivateImplementationDetails>'::CEB73749C899693706EDE1E30C9929B3FD5DD926163831C2FB8BD41E6EFB1126
IL_0005: ldc.i4.4
IL_0006: newobj instance void valuetype [System.Runtime]System.ReadOnlySpan`1<uint8>::.ctor(void*, int32)
// AppendFake(new byte[4] { 116, 101, 115, 116 });
IL_000b: call void TestClass::AppendFake(valuetype [System.Runtime]System.ReadOnlySpan`1<uint8>)
```

So, the array is staticized into a global byte array variable with name "CEB73749C899693706EDE1E30C9929B3FD5DD926163831C2FB8BD41E6EFB1126" inside the hidden class <PrivateImplementationDetails>.
Everytime we use the same u8 string it will reuse the same static global variable and construct only a new ReadOnlySpan that reference it, like these statements:
```cs
AppendFake("test"u8);
AppendFake("test"u8);
```

That emits these:
```
IL_0000: ldsflda valuetype '<PrivateImplementationDetails>'/'__StaticArrayInitTypeSize=5' '<PrivateImplementationDetails>'::CEB73749C899693706EDE1E30C9929B3FD5DD926163831C2FB8BD41E6EFB1126
IL_0005: ldc.i4.4
IL_0006: newobj instance void valuetype [System.Runtime]System.ReadOnlySpan`1<uint8>::.ctor(void*, int32)
// AppendFake(new byte[4] { 116, 101, 115, 116 });
IL_000b: call void TestClass::AppendFake(valuetype [System.Runtime]System.ReadOnlySpan`1<uint8>)
// AppendFake(new byte[4] { 116, 101, 115, 116 });
IL_0010: ldsflda valuetype '<PrivateImplementationDetails>'/'__StaticArrayInitTypeSize=5' '<PrivateImplementationDetails>'::CEB73749C899693706EDE1E30C9929B3FD5DD926163831C2FB8BD41E6EFB1126
IL_0015: ldc.i4.4
IL_0016: newobj instance void valuetype [System.Runtime]System.ReadOnlySpan`1<uint8>::.ctor(void*, int32)
IL_001b: call void TestClass::AppendFake(valuetype [System.Runtime]System.ReadOnlySpan`1<uint8>)
```

kamronbatman:
Before we can merge this, I think we can simplify the concepts. Our goal is:
1. Static Layout, Static Strings
2. Static Layout, Dynamic Strings

The first with static strings can use a dynamic strings builder, but also cache the strings themselves.
The difficult part will be creating a way to specify the strings for an index, I think we decided to use a Dictionary<int, int>.

Projects/UOContent/Network/Packets/IncomingPlayerPackets.cs
```diff
@@ -365,37 +365,37 @@ public static void DisplayGumpResponse(NetState state, SpanReader reader)
...
-                state.Mobile?.SendMessage("Invalid gump response.");
-
-                // state.Disconnect("Invalid gump response.");
-                return;
-            }
+            // var buttonExists = buttonId == 0; // 0 is always 'close'
+            //
+            // if (!buttonExists)
+            // {
+            //     foreach (var e in gump.Entries)
```
kamronbatman, line 372: Should we even bother recording buttons? I think the gump's OnResponse should properly handle invalid input anyways.

PR #1654, opened 2024-01-05
Fix hues for Bags of Sending

blendax78:
You couldn't actually set the hue (randomly/automatically or manually) on item creation.

### DIFF
--- a/Projects/UOContent/Items/Special/Solen Items/BagOfSending.cs
+++ b/Projects/UOContent/Items/Special/Solen Items/BagOfSending.cs
@@ -30,7 +30,7 @@ public partial class BagOfSending : Item, TranslocationItem
     {
         Weight = 2.0;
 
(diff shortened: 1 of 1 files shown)


kamronbatman:
What happens when you try to do something like:

`[add bagofsending yellow`

Or without the argument?

blendax78:
Without fix:
image
Commands:
`[add bagofholding`
`[add bagofholding`
`[add bagofholding`
`[add bagofholding yellow`
`[add bagofholding blue`
`[add bagofholding red`

blendax78:
With fix:
image
Commands:
[add bagofholding
[add bagofholding
[add bagofholding
[add bagofholding yellow
[add bagofholding blue
[add bagofholding red

blendax78:
oops .. meant to say `[add bagofsending` NOT `[add bagofholding` 😆

PR #1660, opened 2024-01-18
feat: Moves TcpServer to another thread. Rewrites Firewall

kamronbatman:
## Breaking Changes
* The Firewall and IP Limiter have been rewritten. Please read the notes carefully!
* `TcpServer.Instances` moved back to `NetState.Instances` - sorry - it was stupid to move it to begin with.

> [!Note]
> Sockets that fail the IP Limiter or Firewall will be immediately and forcibly disconnected.
> This means they will be stuck at "Verifying account..." if it was a real client.

### Summary
- Removes firewall wildcard support.
- Removes `AccessRestrictions`.
- Moves Firewall/IPLimiter to the core.
- Moves `TcpServer` to its own thread.
- Removes the `SocketConnect` and `SocketDisconnect` event sinks.
- Moves `Instances` back to `NetState.Instances`.
- Fixes a long standing bug with bad handling of duplicate listener addresses.

#### Firewall
The firewall has been completely rewritten. There is now an "Admin Firewall" which saves to the config file. Secondarily, there is an internal firewall used exclusively by the TcpServer while processing sockets. The Admin firewall mirrors it's additions/deletions to the internal firewall by adding requests to a queue.

> [!IMPORTANT]  
> **Wildcard firewall entries, such as `X`, `*`, `?` are not allowed.**
> **Ranges in between IP classes or sextets are not allowed.**
> **Please make sure to use one of the following:**
> * IP Address - `192.168.1.1`
> * CIDR - `192.168.1.0/24`
> * Range - `192.168.1.1-192.168.1.100`

#### IP Limiter
The IP Limiter has been completely rewritten. The available configurations are:
```json
"ipLimiter.enable": "True",
"ipLimiter.maxConnectionsPerIP": 10,
"ipLimiter.clearConnectionAttemptsDuration": "00:00:00:10",
"ipLimiter.clearThrottledDuration": "00:00:02:00",
```

The IP Limiter is set up to prevent spamming connections from the same IP. Every time an IP connects, it is added to a connection list. After 10 attempts, the IP is added to the throttle list. To keep the system fast, the connection list is entirely wiped every 10 seconds, and the throttle list is entirely wiped every 2 minutes.

### TODOs
- ~Move PingServer to it's own thread.~
- Add unit tests for the firewall changes including, but not limited to:
  - ~IPAddress to UInt128~
  - ~UInt128 to IPAddress~
  - ~Does blocking actually work the way we expect~
  - ~Verify firewall.cfg is not jacked up and can deserialize/serialize properly~
- Build some benchmarks. Did we make this worse somehow?
- ~Remove the old Utility IP/CIDR code that that shouldn't be used.~
- ~Remove `IPv4ToAddress` and rewrite `MatchClassC`~
- Find a solution for ShardPoller and other IP related systems that need better IPv6 support.
image


Projects/UOContent.Tests/Tests/Firewall/FirewallEntryTests.cs
```diff
@@ -2,19 +2,9 @@
...
-        [InlineData("192.168.*.100-200", typeof(Firewall.WildcardIPFirewallEntry), "192.168.30.150")]
-        [InlineData("::1234:*:1000-1234", typeof(Firewall.WildcardIPFirewallEntry), "::1234:5678:1150")]
-        public void TestFirewallEntry(string entry, Type entryType, string ipToMatch)
-        {
-            var firewallEntry = Firewall.ToFirewallEntry(entry);
-            Assert.IsType(entryType, firewallEntry);
-            Assert.True(firewallEntry.IsBlocked(IPAddress.Parse(ipToMatch)));
-        }
-    }
+
```
kamronbatman, line 9: BAD KAMRON! - ADD TESTS! 😠

Projects/Server/Utilities/Utility.cs
```diff
@@ -158,408 +159,145 @@ public static void Intern(ref IPAddress ipAddress)
...
-                    }
+        destination[..8].Clear(); // Local init is off
+        BinaryPrimitives.WriteUInt32BigEndian(destination.Slice(8, 4), 0xFFFF);
+        ipAddress.TryWriteBytes(destination.Slice(12, 4), out _);
+    }
 
-                    byteIndex += 2;
-                } while (byteIndex < zeroEnd);
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static bool MatchesClassC(this IPAddress ip1, IPAddress ip2) => ip1.MatchCidr(24, ip2);
```
kamronbatman, line 284: Fix naming so it is consistent. Are we using `Match`, or `Matches`?

PR #1662, opened 2024-01-22
fix: Fixes issue with duping items and serialization, and sector lists.

kamronbatman:
> [!Important]
> **This fixes a major issue with [dupe and serialization!**

### Summary

- Fixes [dupe sharing the same SaveBuffer, causing serialization clobbering.
- Fixes an issue with Mobile.SetLocation() -> UpdateRegion() -> OnRegionChange() -> having side effects that makes calling Map.OnMove(0 later in the SetLocation invalid.
- Fixes disabling AutoSave not working.

### DIFF
--- a/Projects/Server/Items/Item.cs
+++ b/Projects/Server/Items/Item.cs
@@ -599,18 +599,16 @@ public class Item : IHued, IComparable<Item>, ISpawnable, IObjectPropertyListEnt
 
             var oldParent = m_Parent;
 
+            if (m_Map != null && oldParent == null && value != null)
+            {
+                m_Map.OnLeave(this);
+            }
+
             m_Parent = value;
 
-            if (m_Map != null)
+            if (m_Map != null && oldParent != null && value == null)
             {
-                if (oldParent != null && m_Parent == null)
-                {
-                    m_Map.OnEnter(this);
-                }
-                else if (m_Parent != null)
-                {
-                    m_Map.OnLeave(this);
-                }
+                m_Map.OnEnter(this);
             }
         }
     }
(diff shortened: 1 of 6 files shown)


Projects/Server/Items/Item.cs
```diff
@@ -599,18 +599,16 @@ public IEntity Parent
...
+            {
+                m_Map.OnLeave(this);
+            }
+
             m_Parent = value;
 
-            if (m_Map != null)
+            if (m_Map != null && oldParent != null && value == null)
             {
-                if (oldParent != null && m_Parent == null)
```
kamronbatman, line 606: This logic from RunUO is completely wrong.

Projects/Server/Items/Item.cs
```diff
@@ -1072,7 +1078,10 @@ public void MoveToWorld(Point3D location, Map map)
 
             if (m_Map != null)
             {
-                m_Map.OnLeave(this);
+                if (!hadParent)
+                {
+                    m_Map.OnLeave(this);
```
kamronbatman, line 1083: We shouldn't try to leave if there was already a parent, because that would be duplicative. This doesn't crash as duplicative cause of the broken logic in setting the Parent.

Projects/Server/Items/Item.cs
```diff
@@ -599,18 +599,16 @@ public IEntity Parent
...
             {
-                if (oldParent != null && m_Parent == null)
-                {
-                    m_Map.OnEnter(this);
-                }
-                else if (m_Parent != null)
-                {
-                    m_Map.OnLeave(this);
-                }
+                m_Map.OnEnter(this);
```
kamronbatman, line 611: When an item is lifted, it is effected moved to the Map's Sector0 list, even though it's location is never set to 0,0,0.
TODO: Tighten up the movement logistics. This shouldn't be allowed.

PR #1666, opened 2024-01-30
fix: Fixes loot with zero amount

kamronbatman:
### Summary

* Adds logging if Amount goes to zero. It should never go to zero!
* Deletes an item if it mutates with zero amount.
* Eliminates possible allocation from `params` for `Type[][]` loot pack construction.
* Renames `Loot.ChestOfHeirloomsContains` to `Loot.RandomChestOfHeirloomsContent`.

### DIFF
--- a/Projects/UOContent/Misc/LootPack.cs
+++ b/Projects/UOContent/Misc/LootPack.cs
@@ -7,88 +7,55 @@ namespace Server
 {
     public class LootPack
     {
-        public static readonly LootPackItem[] Gold =
-        {
-            new(typeof(Gold), 1)
-        };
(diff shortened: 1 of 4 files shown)


Projects/UOContent/Misc/LootPack.cs
```diff
@@ -741,131 +714,140 @@ private int GetRandomOldBonus()
...
-                }
+            if (item.Stackable)
+            {
+                var newAmount = Quantity.Roll();
 
-                if (item.Stackable)
+                if (newAmount <= 0)
                 {
-                    item.Amount = Quantity.Roll();
+                    item.Delete();
```
kamronbatman, line 846: This is the relevant change. Delete the item if it has an invalid amount.

PR #1695, opened 2024-03-05
fix: Reverts to RunUO speeds, fixes direction glitching, adds movement speed interpolation

kamronbatman:
> [!IMPORTANT]  
> Please read through these changes, as they changed certain expectations for how the internal AI thinking/movement work.
> Note that some mobs still don't have smooth movement on ClassicUO due to how the client handles animations/movement.

### Summary
- **Important Change**: Mobs will now move at a more regular pace. If the OnThink results in a move, but the cooldown would otherwise prevent the move, then the movement is scheduled on another timer.
- Added a 400ms delay to mobs turning to face a player to attack in order to avoid glitching between moving and turning.
- Reverted AI thinking speeds back to RunUO specific speeds.
- Reverted the AI thinking to moving conversion delays back to RunUO.
- For thinking speeds that are not exactly the predefined speeds from RunUO, there is a new calculation to determine the correct conversion to movement speed. This stops speeds like `0.35` from being faster than `0.3`

> [!NOTE]  
> **Developer Note**
> BaseAI.CheckMove() no longer contains the check for whether or not a mob is on movement cooldown. This function can now be overwritten without causing issues to figuring out that cooldown.

### DIFF
--- a/Projects/UOContent/Mobiles/AI/BaseAI.cs
+++ b/Projects/UOContent/Mobiles/AI/BaseAI.cs
@@ -2120,7 +2120,7 @@ public abstract class BaseAI
             // When bad state is ok, we can still move if the time until move is less than the fuzzy time
             if (badStateOk)
             {
-                new AIMovementTimer(this, d, TimeSpan.FromMilliseconds(timeUntilMove)).Start();
+                AIMovementTimerPool.GetTimer(TimeSpan.FromMilliseconds(timeUntilMove), this, d).Start();
             }
 
             return MoveResult.BadState;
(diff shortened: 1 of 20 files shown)


Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1992,100 +1988,103 @@ public virtual void WalkRandom(int iChanceToNotMove, int iChanceToDir, int iStep
...
         }
 
-        return thinkingSpeed;
+        return movementSpeed;
     }
 
-    public virtual bool CheckMove() => Core.TickCount - NextMove >= 0;
+    private long TimeUntilNextMove() => Math.Max(NextMove - Core.TickCount, 0);
+
+    public virtual bool CheckMove() => true;
```
kamronbatman, line 2087: We are _removing_ the time until check from `CheckMove` so we can use it to schedule a movement in between "thinks".

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2107,11 +2106,33 @@ public virtual MoveResult DoMoveImpl(Direction d)
             return MoveResult.BadState;
         }
 
+        var delay = (int)(TransformMoveDelay(m_Mobile) * 1000);
+
+        var timeUntilMove = TimeUntilNextMove();
+        if (timeUntilMove > 16)
```
kamronbatman, line 2112: Don't bother if we are within 2 ticks of when we should be able to move.

PR #1699, opened 2024-03-08
fix: Make RelayInfo zero allocation

stefanomerotta:
Removed TextEntry struct and added deferred text response handling for gumps with spans.

The string instantiation is deferred to GetTextEntry method.

### DIFF
--- a/Projects/Server/Buffers/SpanReader.cs
+++ b/Projects/Server/Buffers/SpanReader.cs
@@ -165,12 +165,19 @@ public ref struct SpanReader
         return value;
     }
 
+    /// <summary>
+    /// Check the string in the buffer without converting it with the provided encoding.
+    /// </summary>
+    /// <param name="encoding">Text encoding used for char size calculation</param>
+    /// <param name="fixedLength">Max buffer length to check</param>
+    /// <returns>Byte length of the found string</returns>
+    /// <exception cref="OutOfMemoryException"></exception>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public string ReadString(Encoding encoding, bool safeString = false, int fixedLength = -1)
+    public int SkipString(Encoding encoding, int fixedLength = -1)
     {
         if (fixedLength == 0)
         {
-            return "";
+            return 0;
         }
 
         int byteLength = encoding.GetByteLengthForEncoding();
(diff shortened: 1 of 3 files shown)


Projects/UOContent/Network/Packets/IncomingPlayerPackets.cs
```diff
@@ -420,27 +422,40 @@ public static void DisplayGumpResponse(NetState state, SpanReader reader)
...
 
-            for (var i = 0; i < textEntries.Length; ++i)
+            if (textCount > 0)
             {
-                int entryID = reader.ReadUInt16();
-                int textLength = reader.ReadUInt16();
+                int positionOffset = reader.Position;
 
-                if (textLength > 239)
+                textIds = stackalloc ushort[textCount];
```
kamronbatman, line 433: The counts can be big, and this would allow someone to make a malicious payload to try and crash the server.
Two things, first let's validate the counts coming in and make sure they are equal to what we expect.

Second, we need to allocate an array if the count is  too high, usually it's 1024 bytes. And lastly, we can GC uninitialize the arrays instead of just making them.

kamronbatman (changes requested):

Projects/UOContent/Network/Packets/IncomingPlayerPackets.cs
```diff
@@ -420,27 +422,40 @@ public static void DisplayGumpResponse(NetState state, SpanReader reader)
...
+                        exception.SetStackTrace(new StackTrace());
+                        NetState.TraceException(exception);
+                        state.Mobile?.SendMessage("Invalid gump response.");
+
+                        // state.Disconnect("Invalid gump response.");
+                        return;
+                    }
+
+                    int textStart = reader.Position - positionOffset;
+                    textLength = reader.SkipString(TextEncoding.Unicode, textLength);
```
kamronbatman, line 455: We don't need a special method for this. We can just get the byte count per char from the encoding (it's always 2) and skip 2 × text length by using Seek from Current

kamronbatman (changes requested):

Projects/UOContent/Network/Packets/IncomingPlayerPackets.cs
```diff
@@ -401,11 +404,21 @@ public static void DisplayGumpResponse(NetState state, SpanReader reader)
...
             {
-                switches[i] = reader.ReadInt32();
+                int byteCount = switchCount * 4;
+
+                switches = MemoryMarshal.Cast<byte, int>(reader.Buffer.Slice(reader.Position, byteCount));
+                reader.Seek(byteCount, SeekOrigin.Current);
+
+                if (BitConverter.IsLittleEndian)
+                {
+                    Span<int> reverseSwitches = stackalloc int[switchCount];
```
kamronbatman, line 418: Let's not use stackalloc (or at least have a conditional) when it is above 1024bytes, cause then someone could crash the server. I think the total stack limit is 1MB.

Projects/Server/Buffers/SpanReader.cs
```diff
@@ -165,12 +165,19 @@ public ulong ReadUInt64()
...
+    /// <summary>
+    /// Check the string in the buffer without converting it with the provided encoding.
+    /// </summary>
+    /// <param name="encoding">Text encoding used for char size calculation</param>
+    /// <param name="fixedLength">Max buffer length to check</param>
+    /// <returns>Byte length of the found string</returns>
+    /// <exception cref="OutOfMemoryException"></exception>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public string ReadString(Encoding encoding, bool safeString = false, int fixedLength = -1)
+    public int SkipString(Encoding encoding, int fixedLength = -1)
```
kamronbatman, line 176: Per our conversation, this should be an internal implementation detail. I am ok with a private method called something like `GetStringLength()`, and we _don't_ update the `Position`, then we can do an actual ReadBytes() -> to string for that length.

For public consumption, let's just use `Seek`, especially since this is UTF-16 and we can assume the byte length.

kamronbatman (approved):

PR #1704, opened 2024-03-09
Fix of new GUMP override method

srosellj:


kamronbatman:
Requesting from support for this to be deleted. :)

PR #1717, opened 2024-04-04
fix: Cleans up code with feeding pets and adds batch coin flips random check

kamronbatman:
### Summary
* Removes unnecessary extra loops in increasing loyalty while feeding pets.
* Adds `Utility.CoinFlips` for batch 50/50 successive calls to RNG
* Removes improper usage of `Utility.RandomSample()` and instead uses `Shuffle()`

> [!NOTE]
> **DEVELOPER NOTE**
> 1. Use `Utility.CoinFlips(amount) or `Utility.CoinFlips(amount, max)` instead of looping and doing `Utility.RandomBool()` multiple times. The new method is up to 60x faster.
> 2. Utility.RandomSample is for sampling unique `indices` and has no guard rails against duplicate values

### DIFF
--- a/Projects/UOContent/Mobiles/BaseCreature.cs
+++ b/Projects/UOContent/Mobiles/BaseCreature.cs
@@ -173,6 +173,8 @@ namespace Server.Mobiles
         }
 
         public const int MaxLoyalty = 100;
+        public const int LoyaltyIncreasePerFood = 10;
+        public const int MaxLoyaltyIncrease = MaxLoyalty / LoyaltyIncreasePerFood;
 
         public const int MaxOwners = 5;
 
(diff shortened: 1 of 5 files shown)


Projects/UOContent/Mobiles/BaseCreature.cs
```diff
@@ -4180,91 +4182,87 @@ public virtual bool CheckFoodPreference(Item fed, FoodType type, Type[] types)
...
-                            StaminaSystem.RegenSteps(this as IHasSteps, stamGain * 4);
-                        }
+                if (Core.SE)
+                {
+                    m_Loyalty = MaxLoyalty;
+                }
+                else if (m_Loyalty < MaxLoyalty)
+                {
+                    // 50% chance to increase 10 loyalty per food
+                    m_Loyalty = Math.Min(MaxLoyalty, Utility.CoinFlips(amount, MaxLoyaltyIncrease) * 10);
```
kamronbatman, line 4219: Bulk of the change here.

PR #1721, opened 2024-04-06
fix: Simplifies create food and localizes. Fixes localization fallback

kamronbatman:


kamronbatman:
I have to undo this because affix is ASCII only. :|

PR #1736, opened 2024-04-22
fix: Improves STArray and SpanWriter with packets

kamronbatman:


Projects/Server/Buffers/SpanWriter.cs
```diff
@@ -438,7 +438,7 @@ public int Seek(int offset, SeekOrigin origin)
             _                  => offset // Begin
         });
 
-        if (newPosition >= _buffer.Length)
+        if (newPosition > _buffer.Length)
```
kamronbatman, line 441: This was causing seeks to the end of the spanwriter (for writing packet lengths), to cause the buffer to grow unnecessarily.

Projects/Server/Buffers/STArrayPool.cs
```diff
@@ -15,7 +15,7 @@ namespace Server.Buffers;
  */
 public class STArrayPool<T> : ArrayPool<T>
 {
-    private const int StackArraySize = 8;
+    private const int StackArraySize = 32;
```
kamronbatman, line 18: In .NET 8, this was changed to 32.

TODO: Refactor this to use the newest .net 9 code.

PR #1744, opened 2024-04-27
feat: Cleans up UOP file handling. Adds automatic bounds.bin generation

kaczy93:
It allows for easy access to art dimensions and real bounds based on the pixel transparency.
Used for dynamically sized gumps and stuff.
Code is rough on the edges since it's almost untouched ServUO code, but System.Drawing.Imaging references are gone.
Tested and gives exactly the same results as original ServUO code.


Projects/Server/Client/ArtData.cs
```diff
@@ -0,0 +1,185 @@
...
+                        yMax = y;
+                    }
+                }
+            }
+        }
+
+        return xMax switch
+        {
+            // If no non-black pixels were found, return an empty rectangle
+            -1 => Rectangle2D.Empty,
```
kamronbatman, line 170: Note that in RunUO/ServUO, this value is `[0, 0, 0xFFFF, 0xFFFF]`. Intentionally different here. if the entire pic is black, there are no bounds, so we can keep it as all 00's. This also allows us to easily detect "empty" or "missing" entries without a validation structure.

Projects/Server/TileMatrix/TileMatrix.cs
```diff
@@ -434,9 +439,45 @@ private unsafe LandTile[] ReadLandBlock(int x, int y)
...
+            total = newTotal;
+        }
+
+        return -1;
+    }
+
+    private void ConvertToMapEntries(FileStream stream)
+    {
+        // Sorting by offset to make seeking faster
+        Array.Sort(_uopMapEntries, (a, b) => a.Offset.CompareTo(b.Offset));
```
kamronbatman, line 465: This is only doable because I don't think the hashed file name order is important. I also don't think those files match the order field in the data block.

kamronbatman:
Todo: Create [GenBounds command.

Projects/Server/TileMatrix/TileMatrix.cs
```diff
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
```
kamronbatman, line 20: I don't think this is needed. I'll need to do a pass to remove unused imports.

PR #1748, opened 2024-04-29
Add: PaganDragonsBlood.cs to reflect the graphic change post Stygian …

bohica89:
…Abyss


kamronbatman:
I have been thinking about this. There is no benefit, and only confusion. We have a folder of PaganReagents, and they are considered `Deco-` which indicates they are rares. I think that is good enough.

Thank you for looking into this though!

PR #1755, opened 2024-05-01
fix: Fixes EvilMage/Lord spawning naked, without hair

bohica89:


Projects/UOContent/Mobiles/Monsters/Humanoid/Magic/EvilMage.cs
```diff
@@ -39,8 +40,17 @@ public EvilMage() : base(AIType.AI_Mage)
...
+            if (Core.UOR)
+            {
+                PackItem(new Robe(Utility.RandomNeutralHue()));
+                PackItem(new Sandals());
+            }
+            else
+            {
+                EquipItem(new Robe(Utility.RandomNeutralHue()));
+                EquipItem(new Sandals());
+            }
```
kamronbatman, line 53: Let's just equip it.

```suggestion
            EquipItem(new Robe(Utility.RandomNeutralHue()));
            EquipItem(new Sandals());
```

Projects/UOContent/Mobiles/Monsters/Humanoid/Magic/EvilMage.cs
```diff
@@ -12,6 +12,7 @@ public EvilMage() : base(AIType.AI_Mage)
             Name = NameList.RandomName("evil mage");
             Title = "the evil mage";
             Body = Core.UOR ? 124 : 0x190;
+            Hue = Core.UOR ? 0 : Race.Human.RandomSkinHue();
```
kamronbatman, line 15:
```suggestion
            if (Core.UO)
            {
                Hue = Race.Human.RandomSkinHue();
            }
```

bohica89: Actually, I think EvilMages and EvilMageLords only had the pale models (0x190) and no hue switching.

kamronbatman: Then let's get rid of the hue change?

PR #1756, opened 2024-05-04
fix: Removes broken OrderedHashSet and adds a simple OrderedSet

kamronbatman:
> [!CAUTION]
> **BREAKING CHANGE**
> Removed `OrderdHashSet` and `PooledOrderedHashSet` due to bugs.

> [!NOTE]
> **Developer Note**
> The OrderedSet is not a full data structure. It is not particularly efficient. Pull Requests are welcome for a better implementation, especially if it ends up supporting `ISet<T>` and `IReadOnlySet<T>`

## Summary

The ordered hash set was buggy. It's kind of painful to implement, so for now, I added a simple `OrderedSet` to suffice for gumps. Please reach out if this causes disruption!


Projects/Server/Collections/OrderedSet.cs
```diff
@@ -0,0 +1,70 @@
...
+
+/// <summary>
+/// A data structure designed for scenarios where the order of insertion is important.
+/// Note: This is not particularly effecient.
+/// </summary>
+/// <typeparam name="T">The type of elements in the set.</typeparam>
+public class OrderedSet<T>
+{
+    private readonly Dictionary<T, int> _dictionary;
+    private readonly List<T> _list;
```
kamronbatman, line 30: I am so ashamed. I also don't care in this case since this is for legacy code.

PR #1790, opened 2024-05-23
fix: Fixes house update range bug

kamronbatman:


Projects/UOContent/Multis/Houses/BaseHouse.cs
```diff
@@ -716,7 +716,9 @@ public virtual int GetAosMaxLockdowns()
             return fromSecures + fromVendors + fromLockdowns + fromMovingCrate;
         }
 
-        public bool InRange(Point2D from, int range)
+        public override bool InRange(Point3D from, int range) => InRange(new Point2D(from), range);
+
+        public override bool InRange(Point2D from, int range)
```
kamronbatman, line 721: Crux of the fix. This method wasn't an override before.

PR #1795, opened 2024-05-27
fix: Adds more NetState dumping for debug

kamronbatman:


Projects/Server/Network/NetState/NetState.cs
```diff
@@ -331,7 +332,10 @@ public SecureTrade FindTrade(Mobile m)
         {
             var trade = Trades[i];
 
-            if (trade.From.Mobile == m || trade.To.Mobile == m)
```
kamronbatman, line 334: More RunUO bugs. 😆

PR #1796, opened 2024-05-27
fix: Fixes stalled connections and infinite throttle

kamronbatman:
> [!Warning]
> **Developer Warning**
> The `PacketThrottle` callback return value is now reversed. `true` indicates the connection is _throttled_.

### Summary
- Fixes an issue where connections get stalled forever
- Fixes an issue where the throttler is not working properly
- Removes account attack limiter
- Rewrites IP limiter
- Removes IP restrictions (they weren't used, and not practical)
- Fixes issue where IP limiter was counting before firewall was blocking.

View without whitespace:
https://github.com/modernuo/ModernUO/pull/1796/files?diff=split&w=1


Projects/UOContent/Accounting/AccountAttackLimiter.cs
```diff
@@ -4,160 +4,120 @@
...
-                if (accessLog.HasExpired)
-                {
-                    m_List.RemoveAt(i--);
-                }
-                else if (accessLog.Address.Equals(ipAddress))
-                {
-                    return accessLog;
-                }
-            }
+        return false;
```
kamronbatman, line 39: Do not throttle no matter what because we have disconnected them.

paezao (dismissed):
LGTM!

PR #1804, opened 2024-05-30
GoGump code optimization

nerun:
This update is not mine, I don't know the author. I got this from Discord, but it makes perfect sense to me. In the author words:

> # Overview
> 
> The GoGump class was updated to comply with the constraints of static readonly fields in C#. Specifically, static readonly fields cannot be assigned to outside of a static constructor or a variable initializer. The changes ensure that these fields are properly initialized.
> 
> # Detailed Changes
>
> ## Added Static Constructor:
> 
>A static constructor was introduced to handle the initialization of static readonly fields.
> 
> Static constructors are special constructors that are called automatically before any static members are accessed and are ideal for initializing static fields.
> 
> ## Initialization of Static Readonly Fields:
>
> The fields Felucca, Trammel, Ilshenar, Malas, Tokuno, and TerMur are static readonly. These fields represent different LocationTree instances.
> 
> Previously, these fields were being assigned values within the DisplayTo method, which caused compilation errors because static readonly fields can only be assigned in a static constructor or variable initializer.
> 
> The assignments were moved to the static constructor to comply with the language rules.
>
> ## Removed In-Method Assignments:
> 
>Assignments within the DisplayTo method were removed because they were moved to the static constructor.
>
> The DisplayTo method now simply checks the current map and sets the appropriate LocationTree without attempting to assign values to the static readonly fields.

Additionally, this update also allows the famous _Staff Runebook_ script to work.

### DIFF
--- a/Projects/UOContent/Gumps/Go/GoGump.cs
+++ b/Projects/UOContent/Gumps/Go/GoGump.cs
@@ -7,12 +7,12 @@ namespace Server.Gumps
 {
     public class GoGump : Gump
     {
-        private static LocationTree Felucca;
-        private static LocationTree Trammel;
-        private static LocationTree Ilshenar;
-        private static LocationTree Malas;
-        private static LocationTree Tokuno;
-        private static LocationTree TerMur;
+        public static readonly LocationTree Felucca;
+        public static readonly LocationTree Trammel;
+        public static readonly LocationTree Ilshenar;
+        public static readonly LocationTree Malas;
+        public static readonly LocationTree Tokuno;
+        public static readonly LocationTree TerMur;
 
         private static readonly int PrevLabelOffsetX = PrevWidth + 1;
         private static readonly int PrevLabelOffsetY = 0;
@@ -31,6 +31,16 @@ namespace Server.Gumps
         private readonly int m_Page;
 
         private readonly LocationTree m_Tree;
+
+        static GoGump()
+        {
+            Felucca = new LocationTree("felucca", Map.Felucca);
+            Trammel = new LocationTree("trammel", Map.Trammel);
+            Ilshenar = new LocationTree("ilshenar", Map.Ilshenar);
+            Malas = new LocationTree("malas", Map.Malas);
+            Tokuno = new LocationTree("tokuno", Map.Tokuno);
+            TerMur = new LocationTree("termur", Map.TerMur);
+        }
 
         private GoGump(int page, Mobile from, LocationTree tree, GoCategory node) : base(50, 50)
         {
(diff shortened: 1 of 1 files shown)


Voxpire:
I would recommend implementing these fields as properties using the read only auto pattern; 
`public static object Object { get; }`
The original private fields were being lazy loaded, so the least intrusive change to the system would be to use a private setter;
`public static object Object { get; private set; }`

Projects/UOContent/Gumps/Go/GoGump.cs
```diff
@@ -32,6 +32,16 @@ public class GoGump : Gump
 
         private readonly LocationTree m_Tree;
 
+        static GoGump()
+        {
+            Felucca = new LocationTree("felucca", Map.Felucca);
```
kamronbatman, line 37: We you can assign it to the field directly and eliminate the static constructor.

nerun: You want to say write `GoGump()` instead of `static GoGump()`?

nerun: The way it was before?
```
if (from.Map == Map.Ilshenar)
            {
                tree = Ilshenar ??= new LocationTree("ilshenar", Map.Ilshenar);
            }
```

Projects/UOContent/Gumps/Go/GoGump.cs
```diff
@@ -7,12 +7,12 @@ namespace Server.Gumps
 {
     public class GoGump : Gump
     {
-        private static LocationTree Felucca;
-        private static LocationTree Trammel;
-        private static LocationTree Ilshenar;
-        private static LocationTree Malas;
-        private static LocationTree Tokuno;
-        private static LocationTree TerMur;
+        public static readonly LocationTree Felucca;
```
kamronbatman, line 10: I meant to do it here, but for all maps. Then we can delete the static ctor.
```suggestion
        Felucca = new LocationTree("felucca", Map.Felucca);
```

kamronbatman:
### TODO
We want to make this, and other map related functionality more data driven so that customizing your maps doesn't require gutting code in several places.

Projects/UOContent/Gumps/Go/GoGump.cs
nerun: Aaaah! Ok done.

kamronbatman:
@nerun - One last change. If the LocationTree is used outside of this script, maybe we should move the logic for determining which tree to it's own function. Something like `GetLocationTree(Map map)`?

This brings me to the next proposed change (optional for this PR). Should we move the LocationTree concept to the map directly, so it is defined outside of a Gump?

This way the API would be:

```cs
map.LocationTree
```

Instead of the if/else-if to find the right one referenced by static constants.

Thoughts?

PR #1809, opened 2024-05-31
fix: Console Input is disabled on Linux when no standard input is attached.

kamronbatman:


nerun:
Thank you, i will try later.

PR #1811, opened 2024-05-31
fix: Fixes dupe property copying. Adds IgnoreDupe

kamronbatman:
### Changes
- Adds `[IgnoreDupe]` and `[SerializedIgnoreDupe]`
- Updates all _known_ classes that need the attribute. Some might be missing, please helps us find them!
- Adds `Item.Dupe()` command and encapsulates `CopyProperties` and `OnAfterDuped`. This is also overridable.
- Updates Dupe command to use the new logic.
- Fixes duping multiple kinds of objects that used to be outright broken.

### Bug Fixes
- Fixes issue with durability after duping
- Fixes issue with hue after duping

> [!Note]
> **Developer Note**
> Customizing how duping an item works now requires two steps:
> 1. Add `[IgnoreDupe]` or `[SerializedIgnoreDupe]` to the property/field
> 2. Add custom logic in an `OnAfterDuped` override
>
> When do you need to do this?
> *When the property being copied is not a primitive, and you need to manually deep-clone the contents of the property such as with Lists, Dictionaries, or sub classes.*

### DIFF
--- a/Projects/UOContent/Items/Skill Items/Magical/Spellbook.cs
+++ b/Projects/UOContent/Items/Skill Items/Magical/Spellbook.cs
@@ -96,10 +96,12 @@ public partial class Spellbook : Item, ICraftable, ISlayer, IAosItem
     [SerializedCommandProperty(AccessLevel.GameMaster)]
     private SlayerName _slayer2;
 
+    [SerializedIgnoreDupe]
     [SerializableField(5, setter: "private")]
     [SerializedCommandProperty(AccessLevel.GameMaster, canModify: true)]
     private AosAttributes _attributes;
 
+    [SerializedIgnoreDupe]
     [SerializableField(6, setter: "private")]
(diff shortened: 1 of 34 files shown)


Projects/UOContent/Items/Skill Items/Magical/Spellbook.cs
```diff
@@ -539,13 +541,6 @@ public override void OnAfterDuped(Item newItem)
 
         book.Attributes = new AosAttributes(newItem, _attributes);
         book.SkillBonuses = new AosSkillBonuses(newItem, _skillBonuses);
-        book.Content = _content;
-        book.SpellCount = _spellCount;
-        book.Slayer = _slayer;
-        book.Slayer2 = _slayer2;
-        book.Quality = _quality;
-        book.EngravedText = _engravedText;
-        book.Crafter = _crafter;
```
kamronbatman, line 548: Removed primitives that were copied afterwards that didn't need to be.

PR #1816, opened 2024-06-02
fix: Fixes duping containers and removes copying private setter properties

kamronbatman:
### Summary
- Removes copying private setters
- Fixes duping containers
- Adds public `Dupe.DoDupe` functions for external scripts to hook into the existing logic.

### DIFF
--- a/Projects/UOContent/Commands/Dupe.cs
+++ b/Projects/UOContent/Commands/Dupe.cs
@@ -112,7 +112,9 @@ public static class Dupe
             {
                 for (var j = newItem.Items.Count - 1; j >= 0; j--)
                 {
-                    newItem.RemoveItem(newItem.Items[j]);
+                    var itemToDelete = newItem.Items[j];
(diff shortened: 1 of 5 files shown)


Projects/UOContent/Commands/Dupe.cs
```diff
@@ -40,21 +41,108 @@ private static void DupeInBag_OnCommand(CommandEventArgs e)
...
+                    $"{from.AccessLevel} {CommandLogging.Format(from)} duped {CommandLogging.Format(src)} creating {CommandLogging.Format(newItem)}"
+                );
+            }
+
+            // Recurse for items that have items
+            if (newItem.Items.Count > 0)
+            {
+                for (var j = newItem.Items.Count - 1; j >= 0; j--)
+                {
+                    newItem.RemoveItem(newItem.Items[j]);
```
kamronbatman, line 115: Need to delete the item after this.

PR #1817, opened 2024-06-03
AdminGump: added option 'Do everything'.

nerun:
Added option to generated everything at once, in one single click. (Button ID 114)

admingump


kamronbatman:
Before merging, I am gonna try and think if there is a way to make this DRY.

kamronbatman (approved):

PR #1839, opened 2024-06-17
feat: Standardizes South/East selection gumps

kamronbatman:
### Summary
- Standardizes East/South addon selection gumps by creating an abstract `SelectAddonDirectionGump` static gump.


kamronbatman:
Going to generalize this even more before merging.

PR #1851, opened 2024-07-03
fix: Streamlines potion drink effect

kamronbatman:


Projects/UOContent/Items/Skill Items/Magical/Potions/BasePotion.cs
```diff
@@ -174,14 +174,15 @@ private void Deserialize(IGenericReader reader, int version)
...
-    public static void PlayDrinkEffect(Mobile m)
+    public void PlayDrinkEffect(Mobile m)
     {
         m.RevealingAction();
 
         m.PlaySound(0x2D6);
 
         if (!DuelContext.IsFreeConsume(m))
         {
+            Consume();
```
kamronbatman, line 185: I don't like side effects, but we already do it. So I am cleaning this up.

PR #1870, opened 2024-07-13
fix: Magic NPCs now use spell range

kamronbatman:


Projects/UOContent/Mobiles/BaseCreature.cs
```diff
@@ -350,15 +349,10 @@ public enum TeachResult
         public BaseCreature(
             AIType ai,
             FightMode mode = FightMode.Closest,
-            int iRangePerception = 10,
+            int iRangePerception = DefaultRangePerception,
```
mdodkins, line 352: 👍

mdodkins (approved):
Much cleaner! Looks good to me

PR #1887, opened 2024-07-24
Fix consume type when UseAllRess

raydienull:


Voxpire (dismissed):

kamronbatman:
Upon research, I am not sure if this is correct. Do we have any sources for this change?

PR #1900, opened 2024-08-05
fix: Hair and facial hair "teleporting" when mobile dies several times

mdodkins:
* Added World.NewVirtual for creating virtual serial numbers
* Reserved range 0x7EEEEEEE to 0x7FFFFFFF for virtual serials
* Hair and Facial hair (for mobiles) now use virtual serials instead of FakeSerial() functions
* Added CorpseHair and CorpseFacialHair, serializable types which Corpses create based on the virtual hair and facial hair of the mobile that died

Corpse hair and facial hair now persists across save/load and hair and facial hair no longer teleport to newest corpse.

### DIFF
--- a/Projects/Server/Mobiles/Mobile.cs
+++ b/Projects/Server/Mobiles/Mobile.cs
@@ -187,7 +187,7 @@ public delegate bool AllowBeneficialHandler(Mobile from, Mobile target);
 public delegate bool AllowHarmfulHandler(Mobile from, Mobile target);
 
 public delegate Container CreateCorpseHandler(
-    Mobile from, HairInfo hair, FacialHairInfo facialhair, List<Item> initialContent, List<Item> equippedItems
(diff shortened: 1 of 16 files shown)


Projects/Server/Mobiles/Mobile.cs
```diff
@@ -4764,19 +4772,7 @@ public virtual void Kill()
...
-        }
-
-        FacialHairInfo facialhair = null;
-        if (m_FacialHair != null)
-        {
-            facialhair = new FacialHairInfo(m_FacialHair.ItemID, m_FacialHair.Hue);
-        }
-
-        var c = CreateCorpseHandler?.Invoke(this, hair, facialhair, content, equip);
+        var c = CreateCorpseHandler?.Invoke(this, Hair, FacialHair, content, equip);
```
mdodkins, line 4775: @kamronbatman is this ok? or is it potentially a problem to not be re-creating the objects before passing them like this?

mdodkins: I mean... it works.... but I wondered if there was a reason it had been done like this, related to the use of Invoke :)

PR #1904, opened 2024-08-06
fix: Converts ResurrectGump to Static/Dynamic Gumps

IcculusC:
Looked into migrating the `ResurrectionGump`.

It turns out that it covers both the regular use and a special RunUO gump for the `PricedHealer`.  Therefore I split it into two new `StaticGump` implementations.  They share much of the same logic in their response handler, and I'm open to suggestions for how to abstract that out.

### DIFF
--- a/Projects/UOContent/Gumps/PricedResurrectGump.cs
+++ b/Projects/UOContent/Gumps/PricedResurrectGump.cs
@@ -1,6 +1,5 @@
 using System;
 using Server.Engines.Virtues;
-using Server.Items;
 using Server.Misc;
 using Server.Mobiles;
 using Server.Network;
@@ -23,7 +22,7 @@ public class PricedResurrectGump : StaticGump<PricedResurrectGump>
     {
         builder.SetNoClose();
 
-        builder.AddPage(0);
+        builder.AddPage();
 
         builder.AddImage(0, 0, 3600);
 
--- a/Projects/UOContent/Gumps/PricedResurrectGump.cs
+++ b/Projects/UOContent/Gumps/PricedResurrectGump.cs
@@ -1,0 +1,179 @@
+using Server.Engines.Virtues;
+using Server.Misc;
+using Server.Mobiles;
+using Server.Network;
+
+namespace Server.Gumps
+{
+    public class PricedResurrectGump : StaticGump<PricedResurrectGump>
+    {
+        private readonly Mobile m_Healer;
+        private readonly int m_Price;
+
+        public PricedResurrectGump(Mobile owner, Mobile healer, int price)
+            : base(150, 50)
+        {
+            m_Healer = healer;
+            m_Price = price;
+        }
+
+        protected override void BuildLayout(ref StaticGumpBuilder builder)
+        {
+            builder.SetNoClose();
+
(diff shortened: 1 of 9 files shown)


Projects/UOContent/Gumps/PricedResurrectGump.cs
```diff
@@ -0,0 +1,178 @@
...
+            var loss = (100.0 - (4.0 + player.ShortTermMurders / 5.0)) / 100.0; // 5 to 15% loss
+
+            if (loss < 0.85)
+            {
+                loss = 0.85;
+            }
+            else if (loss > 0.95)
+            {
+                loss = 0.95;
+            }
```
kamronbatman, line 152:
```suggestion
            var loss = Math.Clamp((100.0 - (4.0 + player.ShortTermMurders / 5.0)) / 100.0, 0.85, 0.95); // 5 to 15% loss
```

Projects/UOContent/Gumps/PricedResurrectGump.cs
```diff
@@ -0,0 +1,178 @@
...
+        builder.AddImageTiled(394, 0, 1, 217, 9155);
+        builder.AddImageTiled(0, 216, 395, 1, 9157);
+        builder.AddImageTiled(0, 0, 1, 217, 9155);
+    }
+
+    public override void OnResponse(NetState state, in RelayInfo info)
+    {
+        var from = state.Mobile;
+
+        from.CloseGump<PricedResurrectGump>();
```
kamronbatman, line 78: Where was this before? Because It is effectively closed at this point. Do we want to add it to the SendTo override instead?

IcculusC: good call, it was in the top of the handler, I considered that as well, but it's been a while...

Projects/UOContent/Gumps/ResurrectGump.cs
```diff
@@ -4,258 +4,177 @@
...
+    }
 
-        public ResurrectGump(Mobile owner, ResurrectMessage msg) : this(owner, owner, msg)
-        {
-        }
+    public ResurrectGump(
+        Mobile owner, Mobile healer, ResurrectMessage msg = ResurrectMessage.Generic,
+        bool fromSacrifice = false, double hitsScalar = 0.0
+    )
+        : base(100, 0)
```
kamronbatman, line 42:
```suggestion
    ) : base(100, 0)
```

Projects/UOContent/Gumps/ResurrectGump.cs
```diff
@@ -4,258 +4,177 @@
...
-            AddButton(200, 227, 4005, 4007, 0);
-            AddHtmlLocalized(235, 230, 110, 35, 1011012); // CANCEL
+    public override void OnResponse(NetState state, in RelayInfo info)
+    {
+        var from = state.Mobile;
 
-            AddButton(65, 227, 4005, 4007, 1);
-            AddHtmlLocalized(100, 230, 110, 35, 1011011); // CONTINUE
-        }
+        from.CloseGump<ResurrectGump>();
```
kamronbatman, line 75: Same as above.

PR #1915, opened 2024-08-09
fix: migrate RunebookGump to DynamicGump

IcculusC:
migrated `RunebookGump` to `DynamicGump`, cleaned up some of the repeated logic, reduced nesting

### DIFF
--- a/Projects/UOContent/Gumps/RunebookGump.cs
+++ b/Projects/UOContent/Gumps/RunebookGump.cs
@@ -7,215 +7,136 @@ using Server.Spells.Chivalry;
 using Server.Spells.Fourth;
 using Server.Spells.Seventh;
 
-namespace Server.Gumps
+namespace Server.Gumps;
+
+public class RunebookGump : DynamicGump
 {
-    public class RunebookGump : Gump
+    private readonly Runebook _book;
+    public Runebook Book => _book;
+
+    public RunebookGump(Runebook book) : base(150, 200) => _book = book;
+
+    protected override void BuildLayout(ref DynamicGumpBuilder builder)
     {
-        public RunebookGump(Mobile from, Runebook book) : base(150, 200)
+        AddBackground(ref builder);
+        AddIndex(ref builder);
+
+        for (var page = 0; page < 8; ++page)
         {
-            Book = book;
+            builder.AddPage(2 + page);
 
-            AddBackground();
-            AddIndex();
+            builder.AddButton(125, 14, 2205, 2205, 0, GumpButtonType.Page, 1 + page);
 
-            for (var page = 0; page < 8; ++page)
+            if (page < 7)
             {
-                AddPage(2 + page);
+                builder.AddButton(393, 14, 2206, 2206, 0, GumpButtonType.Page, 3 + page);
+            }
 
-                AddButton(125, 14, 2205, 2205, 0, GumpButtonType.Page, 1 + page);
-
-                if (page < 7)
-                {
-                    AddButton(393, 14, 2206, 2206, 0, GumpButtonType.Page, 3 + page);
-                }
-
-                for (var half = 0; half < 2; ++half)
-                {
-                    AddDetails(page * 2 + half, half);
-                }
+            for (var half = 0; half < 2; ++half)
+            {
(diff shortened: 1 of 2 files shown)


Projects/UOContent/Gumps/RunebookGump.cs
```diff
@@ -7,504 +7,448 @@
...
-            for (var page = 0; page < 8; ++page)
-            {
-                AddPage(2 + page);
+namespace Server.Gumps;
 
-                AddButton(125, 14, 2205, 2205, 0, GumpButtonType.Page, 1 + page);
+public class RunebookGump : DynamicGump
+{
+    private readonly Runebook _book;
+    public Runebook Book => _book;
```
IcculusC, line 15: [See this linky link for details on Book](https://github.com/modernuo/ModernUO/pull/1915/files#diff-b59c51b0cd5556d3336f71385690c4b319a649247b39de94ce3a3dac1703a6a5R184)

Projects/UOContent/Gumps/RunebookGump.cs
```diff
@@ -7,504 +7,448 @@
...
+        {
+            buttonID -= 2;
 
-                var index = buttonID / 6;
-                var type = buttonID % 6;
+            var index = buttonID / 6;
+            var type = buttonID % 6;
 
-                if (index >= 0 && index < Book.Entries.Count)
+            if (index >= 0 && index < _book.Entries.Count)
```
kamronbatman, line 259: Invert and return early?

Projects/UOContent/Gumps/RunebookGump.cs
```diff
@@ -7,504 +7,448 @@
...
-                else
-                {
-                    Book.Openers.Remove(from);
+                            break;
+                        }
                 }
             }
+            else
+            {
+                _book.Openers.Remove(from);
```
kamronbatman, line 386: Actually, I think openers is always removed. Is there a case where it isn't?

Projects/UOContent/Gumps/RunebookGump.cs
```diff
@@ -7,504 +7,448 @@
...
+                                    502413,
+                                    null,
+                                    0x35
+                                ); // That cannot be done while the book is locked down.
                             }
-                        case 2: // Set default
+
+                            break;
+                        }
+                    case 2: // Set default
```
kamronbatman, line 311: This case doesn't remove openers. It's actually a bug! I think we should remove openers no matter what.

Projects/UOContent/Items/Skill Items/Magical/Runebook.cs
```diff
@@ -263,7 +263,7 @@ public override void OnDoubleClick(Mobile from)
             }
 
             from.CloseGump<RunebookGump>();
-            from.SendGump(new RunebookGump(from, this));
+            from.SendGump(new RunebookGump(this));
 
             Openers.Add(from);
```
kamronbatman, line 268: Instead of this, we should make a convenience function and call it from OnResponse from situations where the gump should be opened again.

Projects/UOContent/Gumps/RunebookGump.cs
IcculusC: sure thing

Projects/UOContent/Gumps/RunebookGump.cs
IcculusC: it's not removed in the cases where you receive a new `RunebookGump` in return for your response, but I see how I can clean it up

Projects/UOContent/Gumps/RunebookGump.cs
IcculusC: it's not actually a bug, actually - there are two cases where it re-sends the gump and therefore they are still openers

Projects/UOContent/Items/Skill Items/Magical/Runebook.cs
IcculusC: so I'm going to make this a static `Resend` function on the gump so the `InternalPrompt` can call it effectively, lmk if this is dumb

PR #1919, opened 2024-08-10
fix: Fix NetStateGumps and other Gumps cleanup

stefanomerotta:
Fixes wrong behavior of NetStateGumps: added null check to send gump in case of null NetState and removed empty list.

Added gump list max size checks.

Removed unused/unwanted methods.

### DIFF
--- a/Projects/UOContent/Gumps/Base/GumpSystem.cs
+++ b/Projects/UOContent/Gumps/Base/GumpSystem.cs
@@ -26,8 +26,8 @@ namespace Server.Gumps;
 
 public static partial class GumpSystem
 {
-    private const int GumpCap = 512;
     private const int InitialCapacity = 4;
+    public const int GumpCap = 512;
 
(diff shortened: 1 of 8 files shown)


Projects/UOContent/Gumps/Base/GumpSystem.cs
```diff
@@ -161,12 +134,19 @@ private static void Send(NetState ns, BaseGump gump, bool singleton)
 
             if (!replaced)
             {
+                if (list.Count >= GumpCap)
+                {
+                    _logger.Information("Exceeded gump cap, disconnecting...");
+                    ns.Disconnect("Exceeded gump cap.");
```
IcculusC, line 140: Interesting.  I don't have all the context, but I wonder if it would be better to start recycling the gumps on the NetState rather than disconnecting it.  I could foresee a situation where this happens due to something the player user has no control over.  That sounds frustrating.

kamronbatman (approved):

PR #1920, opened 2024-08-10
fix: Moves pricing from PricedHealer to BaseHealer - Combines Res Gumps

IcculusC:


Projects/UOContent/Gumps/PricedResurrectGump.cs
```diff
@@ -7,6 +7,16 @@ namespace Server.Gumps;
 
 public class PricedResurrectGump : StaticGump<PricedResurrectGump>
 {
+    public static void Configure()
```
IcculusC, line 10: I'll remove this, forgot to use `add -p`

Projects/UOContent/Mobiles/Healers/BaseHealer.cs
```diff
@@ -14,8 +14,17 @@ public abstract partial class BaseHealer : BaseVendor
 
     private DateTime _nextResurrect;
 
-    public BaseHealer()
+    [SerializableField(0)] [SerializedCommandProperty(AccessLevel.GameMaster)]
```
IcculusC, line 17: is this how this works? what do I need to do here?  this was how it was serialized in the `PricedHealer` before I moved his stuff here

PR #1924, opened 2024-08-12
Typo in GateTravel sends you to the origin map

Reetus:
Should be the linked gates Map, not TargetMap, as that is the map you're gating from...

Therefore, gating across maps sends you to the correct location on the original map...

Example, gating to Luna bank...
image


kamronbatman (approved):

kamronbatman:
Thanks!

PR #1960, opened 2024-09-21
fix: Ensure server listing address is used when set in config

jkappers:
This is important for container management hosts where the server container doesn't have access to the host IP.

The following scenarios were tested:

| scenario                                                   | public ip | private ip |
|:-----------------------------------------------------------|:---------:|:----------:|
| Docker and client on same host                          | Y         | Y          |
| Docker and client on different hosts, same network      | Y         | Y          |
| Docker and client on different hosts, different network | Y         | N/A        |

### DIFF
--- a/Projects/UOContent/Misc/ServerList.cs
+++ b/Projects/UOContent/Misc/ServerList.cs
@@ -63,7 +63,7 @@ namespace Server.Misc
             {
                 Resolve(Address, out _publicAddress);
 
-                if (_publicAddress is not null)
+                if (_publicAddress != null)
                 {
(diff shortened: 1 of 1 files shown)


kamronbatman (dismissed):

Projects/UOContent/Misc/ServerList.cs
```diff
@@ -60,6 +62,13 @@ public static void Initialize()
             else
             {
                 Resolve(Address, out _publicAddress);
+
+                if (_publicAddress is not null)
```
kamronbatman, line 66:
```suggestion
                if (_publicAddress != null)
```

PR #1970, opened 2024-10-08
fix: Fixes Occlo so it uses the actual addon when decorations are generated.

Bohicatv:
Statics do not have player mobile integrations.


kamronbatman:
Add-ons are traditionally only supposed to be used in a house. Do we know if it was usable on OSI?

Bohicatv:
The original static generates inside of the tailor shop which isn't usable by any player mobile. Changing it to the component fixes this for public use.

PR #1985, opened 2024-10-28
fix: Fixes Timer crash due to negative inputs.

kamronbatman:
### Summary

Fixes a few minor issues with timers:
- Timer.Delay and Timer.Interval was not reflecting the actual tick time (aligned to the next 8ms)
- Negative delay values were causing a crash when DateTime.Now - delay was below DateTime.MinValue
- Timer.Next now reflects the correct wall clock tick time based on the adjusted Delay.
- Timer.Next is not assigned when the timer is started. This was important for timers that were created, but started later.

### DIFF
--- a/Projects/Server/Timer/Timer.TimerWheel.cs
+++ b/Projects/Server/Timer/Timer.TimerWheel.cs
@@ -1,6 +1,6 @@
 /*************************************************************************
  * ModernUO                                                              *
- * Copyright 2019-2023 - ModernUO Development Team                       *
+ * Copyright 2019-2024 - ModernUO Development Team                       *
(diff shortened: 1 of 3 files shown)


Projects/Server/Timer/Timer.TimerWheel.cs
```diff
@@ -214,18 +224,19 @@ private static void AddTimer(Timer timer, long delay)
                     }
                 }
 
+                timer.Next = Core.Now + timer.Delay;
```
kamronbatman, line 227: Note that `Next` is now assigned when the timer is added. It no longer erroneously reflects the Next tick based on when the timer is added.

PR #1990, opened 2024-11-05
fix: Fixes monster ability firing after dying

kamronbatman:


zebmastah:
Code has beed tested on Ubuntuserver. Both Lord Oakes with silvani and semidar has been successfully killed without being able to trigger a server crash.

PR #2012, opened 2024-12-11
fix: Update Hythloth teleport locations in Felucca and Trammel to new coordinates to prevent looping and stuck.

Bohicatv:


kamronbatman:
I assume this is era specific. Can we try to identify which client version this changed in?

Bohicatv:
This isn't era specific. I will always add if it is. However, there was an issue where the middle teleporter wouldn't teleport you, because of a terrain issue with how they designed this area transition, mainly with the stairs, back in 1997. This change moves the teleporters off the stairs to prevent player collision issues, as well as, teleporters being positioned in a way that would cause a loop back and forth.

kamronbatman:
I see, so this may deviate from OSI, that's fine.

kamronbatman (approved):

PR #2019, opened 2024-12-16
Fix HolidayTree not serializing components

Reetus:
Components isn't serialized in V2...

Without the null check in OnAfterDelete, existing HolidayTree's will crash the server when being redeed'd after server restart.


kamronbatman:
@Reetus - I think the idea here (and I am sorry I failed), was to have the addon register itself. Something like `AfterDeserialization` -> `_tree._components.Add(this);` so we don't have to serialize more than necessary.

Thoughts on this approach instead?

kamronbatman:
Going to consolidate with other PRs.

PR #2020, opened 2024-12-16
fix: Fixes overlapping text in TithingGump

Reetus:
X coordinate is wrong for 'Tithe:' text, therefore 'Gold:' and 'Tithe:' display overlapping


kamronbatman:
Thanks! This was a bad copy-pasta!

PR #2021, opened 2024-12-16
Fix missing LabelNumber on DisguiseKit

Reetus:
Would previously display as 'Wig Stand'


kamronbatman:
Looks like I accidentally deleted it. thank you!

PR #2024, opened 2024-12-16
fix: Fix HolidayTree not serializing components

Reetus:


Projects/UOContent/Items/Special/Holiday/Christmas/HolidayTree.cs
```diff
@@ -16,7 +17,7 @@ public partial class HolidayTree : Item, IAddon
     [SerializedCommandProperty(AccessLevel.GameMaster)]
     private Mobile _placer;
 
-    private Item[] _components;
+    private List<Item> _components = [];
```
kamronbatman, line 20: Made this an array originally cause it almost never changes, right?

Projects/UOContent/Items/Special/Holiday/Christmas/HolidayTree.cs
```diff
@@ -208,5 +220,15 @@ public override void OnDoubleClick(Mobile from)
                 _tree.OnDoubleClick(from);
             }
         }
+
+        public static void Deserialize(IGenericReader reader, int version)
+        {
+        }
```
kamronbatman, line 226: This should not be needed, instead we use a MigrateFrom.

I can fix up this PR, thank you!

PR #2033, opened 2024-12-27
fix: Force closes bad connections

kamronbatman:


Projects/Server/Network/TcpServer.cs
```diff
@@ -123,50 +140,131 @@ public static Socket CreateListener(IPEndPoint ipep)
...
+            }
+        }
+    }
+
+    private static async Task<Socket> ValidateSocketAsync(Socket socket)
+    {
+        var firstBytes = new byte[5];
+        try
+        {
+            await socket.ReceiveAsync(firstBytes, SocketFlags.Peek).WaitAsync(TimeSpan.FromMilliseconds(20));
```
kamronbatman, line 189: Timeout so short won't work

Bohicatv: After some testing, 30 second timeout (30000) seems better. You can work your way down from there.

kamronbatman: I would want the server to wait the least amount of time between accept and first packet to minimize threat surface from DDoS.

Bohicatv: You can close bad connections on the network/firewall side by changing the TCP timeouts there to a 1 to 2 second threshold. The thing with TCP, its error checking, so you want the network/firewalls layer to cut off bad connections so MUO isn't doing it.

kamronbatman: I agree that this should definitely be configured outside of the game server. I believe the time out is set by the socket options as well in the code. And by default it never times out. I'm not sure if this overrides the operating system though?

kamronbatman:
This code won't work the way I want it to. I have a version of this that uses another epoll instance (wepoll on Windows), but due to a bug in wepoll, I won't be able to do "clever' things the way I want to.

I may come around to a deep dive into wepoll so I can fix the bug, but I am not set up (nor do I have the requisite knowledge at this time) to figure it out. cc @jaedan if you are interested in fixing the wepoll bug.

Closing this for now and expect users that need to close bad connections to use a proxy in front of MUO.

PR #2034, opened 2024-12-30
NetState: add Task field

MaxKellermann:
This allows making packet handlers async.

### DIFF
--- a/Projects/Server/Network/NetState/NetState.cs
+++ b/Projects/Server/Network/NetState/NetState.cs
@@ -29,6 +29,8 @@ using System.Net.Sockets;
 using System.Network;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Threading; // for CancellationTokenSource
+using System.Threading.Tasks;
 
 namespace Server.Network;
 
@@ -67,6 +69,14 @@ public partial class NetState : IComparable<NetState>, IValueLinkListNode<NetSta
     private long[] _packetThrottles;
     private long[] _packetCounts;
     private string _disconnectReason = string.Empty;
+
+    /**
+     * If this field is not null, then a packet handler is currently
+     * running asynchronously and handling of other packets should be
+     * suspended.
+     */
+    internal Task _packetHandlerTask;
+    internal CancellationTokenSource _packetHandlerCancel;
 
     internal ParserState _parserState = ParserState.AwaitingNextPacket;
     internal ProtocolState _protocolState = ProtocolState.AwaitingSeed;
@@ -533,9 +543,36 @@ public partial class NetState : IComparable<NetState>, IValueLinkListNode<NetSta
         }
     }
 
+    /**
+     * To be called when a packet handler wishes to continue execution
+     * asynchronously.  As long as this #Task runs, no more packets
+     * are handled.
+     */
+    public void AwaitPacketHandlerTask(Task task, CancellationTokenSource cancel)
+    {
+        _packetHandlerTask = task;
+        _packetHandlerCancel = cancel;
+
+        /* schedule the completion callbacks to the main thread */
+        var scheduler = TaskScheduler.FromCurrentSynchronizationContext();
+
+        task.ContinueWith(_ => {
+            _packetHandlerTask = null;
+            _packetHandlerCancel = null;
+        }, cancel.Token, TaskContinuationOptions.OnlyOnRanToCompletion, scheduler);
+
+        task.ContinueWith(t => {
+            _packetHandlerTask = null;
+            _packetHandlerCancel = null;
+
+            Disconnect($"Async packet handler error: {t.Exception.InnerException}");
+            TraceException(t.Exception.InnerException);
+        }, cancel.Token, TaskContinuationOptions.OnlyOnFaulted, scheduler);
+    }
+
     public void HandleReceive(bool throttled = false)
     {
-        if (!_running)
+        if (!_running || _packetHandlerTask != null)
         {
             return;
         }
@@ -550,7 +587,7 @@ public partial class NetState : IComparable<NetState>, IValueLinkListNode<NetSta
         try
         {
             // Process as many packets as we can synchronously
-            while (_running && _parserState != ParserState.Error && _protocolState != ProtocolState.Error)
+            while (_running && _parserState != ParserState.Error && _protocolState != ProtocolState.Error && _packetHandlerTask == null)
             {
                 var buffer = reader.AvailableToRead();
                 var length = buffer.Length;
@@ -1092,6 +1129,12 @@ public partial class NetState : IComparable<NetState>, IValueLinkListNode<NetSta
         }
 
         _running = false;
+
+        /* if an asynchronous packet handler is still running, cancel
+           it */
+        _packetHandlerCancel?.Cancel();
+        _packetHandlerCancel = null;
+        _packetHandlerTask = null;
 
         _disconnectReason = reason;
         _disposed.Enqueue(this);


MaxKellermann:
I can't sign the CLA because the CLA text doesn't specify the exact licensing conditions which I am expected to grant to this project. Please clarify the CLA text.

Projects/Server/Network/NetState/NetState.cs
```diff
@@ -68,6 +70,14 @@ public partial class NetState : IComparable<NetState>, IValueLinkListNode<NetSta
     private long[] _packetCounts;
     private string _disconnectReason = string.Empty;
 
+    /**
+     * If this field is not null, then a packet handler is currently
+     * running asynchronously and handling of other packets should be
+     * suspended.
+     */
+    internal Task _packetHandlerTask;
```
IcculusC, line 78: I may be misunderstanding, but, if this is going to block the handling of future packets, what do we gain by doing packet handling asynchronously in the first place?

MaxKellermann: This blocks handling packets of this client only. What you gain is that the server can continue handling packets from other clients.
Blocking packet handling in the current client is necessary because otherwise you'd be handling packets in random order, and that will cause major pains. All current code assumes that packet handling (within one client) is serialized.

IcculusC: > This blocks handling packets of this client only.

I noticed it's contained to the actual `NetState` of a single connection.  I'm mostly just curious what the use case is, if you're willing to share.  It's worked the same way for 20 years and I haven't ever given it much thought.

so, if I understand this correctly, you want to "block" subsequent packets until the `Task` is done so you can maintain ordering?

MaxKellermann: Simple example: you implement an AccountHandler that does a network request to verify a password. If you do it synchronously, the whole shard freezes until the auth server responds. With this PR, it's possible to do such a network request asynchronously, and let the shard keep running while that one client is authenticating.

> you want to "block" subsequent packets until the `Task` is done so you can maintain ordering?

Yes, that's what I just said. It's not something I "want". It's technically necessary due to how the rest of the code base is implemented.

IcculusC: /shrug

thanks.

nice improvement, in my opinion, "technically necessary" is a stretch, again in my opinion - we await Kamron re: CLA

MaxKellermann: So, in your opinion, the current code base can handle out-of-order handling of packets?
This isn't something one can have an opinion on - it is either true or false. There's no room for opinions and no room for "stretching".

IcculusC: > So, in your opinion, the current code base can handle out-of-order handling of packets?

I never said this.  I also agree it's technically necessary for your example.

MaxKellermann: I know you never said this explicitly, you only said that it "is a stretch", and I was trying to guess what that means. Because this isn't a "stretchy" thing. Only true or false.

IcculusC: all I intended to say was that it's not a technical shortcoming, we don't have/support asynchronous packet handling, shard goes brr either way

it's actually kind of weird that nobody has implemented it in however many years this pattern has been available in C# (assuming not the whole 20~ year life of this codebase)

it's a cool improvement, and it's simple, and I like it

MaxKellermann: > all I intended to say was that it's not a technical shortcoming, we don't have/support asynchronous packet handling

But you replied to something that was *not* about asynchronous packet handling but about blocking packet handling to maintain ordering.
It looks like you mixed up this context, which is the root of our misunderstanding: I did not say async packet handling is necessary. I only said ordered/serial packet handling is technically necessary. And you fully agreed with this. (And really nothing is a stretch.)

IcculusC: 👍

Projects/Server/Network/NetState/NetState.cs
```diff
@@ -533,9 +543,36 @@ private void LogPacket(ReadOnlySpan<byte> buffer, bool incoming)
...
+
+            Disconnect($"Async packet handler error: {t.Exception.InnerException}");
+            TraceException(t.Exception.InnerException);
+        }, cancel.Token, TaskContinuationOptions.OnlyOnFaulted, scheduler);
+    }
+
     public void HandleReceive(bool throttled = false)
     {
-        if (!_running)
+        if (!_running || _packetHandlerTask != null)
```
kamronbatman, line 575: I don't believe we should block (both on L575 and L590) just because a task is executing asynchronously. This could lag the client and potentially desync the client because other packets are not processed in a timely manner.

I agree that processing a packet _out of order_ should be avoided, however to properly handle that requires a more robust, holistic solution such as a higher level state machine that understands the packets, context, and inspects the data coming through the pipe. This way you aren't indiscriminately blocking the socket.

Furthermore, we should have an answer/solution for how we want to prevent a potential DOS vuln by clients that spam packets. It would be a shame if a dev implements this, for example, to push stats to a web server based on incoming packets, and their server locks up.

I understand that the intention of this PR is to handle async account login, however when used for other scenarios we may want more granular control and power to not block/micro-freeze the client.

Projects/Server/Network/NetState/NetState.cs
```diff
@@ -533,9 +543,36 @@ private void LogPacket(ReadOnlySpan<byte> buffer, bool incoming)
         }
     }
 
+    /**
+     * To be called when a packet handler wishes to continue execution
+     * asynchronously.  As long as this #Task runs, no more packets
+     * are handled.
+     */
+    public void AwaitPacketHandlerTask(Task task, CancellationTokenSource cancel)
```
kamronbatman, line 551: What is wrong with async/await and a try/catch? Since the code starts on the game thread, which has a synchronization context that pushes continuations back to the game thread, I don't think this is necessary?

Projects/Server/Network/NetState/NetState.cs
MaxKellermann: How would this lag/desync the client any more than current synchronous execution would lag the client?
I wish the ModernUO-Client code base were more robust, but it's not. You can argue all day over your "holistic solution", but it doesn't exist, will not exist for a long time (if ever), and this PR tries to make the best of the bad code we have. It is a minimal solution.
And it solves a grave problem - you know which one, because you authored it (obviously based on a misunderstanding how .NET async works). I mean, seriously, you argue about client lag, when your bug causes the whole shard to freeze for seconds every time somebody logs in?
May I point out that the way you authored this bug, it contains an auth bypass vulnerability because you did *not* block handling further packets during the (pseudo-)async handler? What was your holistic approach to authentication there?
This PR is strictly an improvement, and you did not claim otherwise; your review only demands to have a perfect solution for everything in a scope wider than this PR, and apparently a strict improvement is not enough for you. This would be a valid discussion if you would judge your own contributions by the same standards, but you don't.

Bohicatv: Because, If the game is checking packets too slowly or missing them because the handler is busy, it could lead to a delay in processing user inputs or server responses. This would certainly result in a piling cascade of desync, even on modern hardware, where the player's actions are not reflected in real-time.

To go even further, this would effectively allow duping without issues as it could be measured as the `_packetHandlerTask` could send multiple packets to account for the lost ones.

kamronbatman: > How would this lag/desync the client any more than current synchronous execution would lag the client? I wish the ModernUO-Client code base were more robust, but it's not. You can argue all day over your "holistic solution", but it doesn't exist, will not exist for a long time (if ever), and this PR tries to make the best of the bad code we have. It is a minimal solution. And it solves a grave problem - you know which one, because you authored it (obviously based on a misunderstanding how .NET async works). I mean, seriously, you argue about client lag, when your bug causes the whole shard to freeze for seconds every time somebody logs in? May I point out that the way you authored this bug, it contains an auth bypass vulnerability because you did _not_ block handling further packets during the (pseudo-)async handler? What was your holistic approach to authentication there? This PR is strictly an improvement, and you did not claim otherwise; your review only demands to have a perfect solution for everything in a scope wider than this PR, and apparently a strict improvement is not enough for you. This would be a valid discussion if you would judge your own contributions by the same standards, but you don't.

Please provide an auth bypass exploit and proof of concept of how the shard can freeze for seconds on end, given nothing but the code in main. Until then, this PR is on hold.

MaxKellermann:
> Please provide an auth bypass exploit and proof of concept of how the shard can freeze for seconds on end, given nothing but the code in main. Until then, this PR is on hold.

What the f*cking sh*t, @kamronbatman! You want to put this PR on hold until I spend my free time to develop an exploit for your auth bypass vulnerability?

We could discuss this vulnerability, I could explain it to you - but you demand that I write an exploit for you?

Hey man, I'm trying to help you with your project. I already spent a considerable amount of time on a piece of shitty code that I've never used and will never use. I left UO some 25 years ago. I just love to code. But I don't like this sh*t. Enough!

kamronbatman:
> > Please provide an auth bypass exploit and proof of concept of how the shard can freeze for seconds on end, given nothing but the code in main. Until then, this PR is on hold.
> 
> What the f_cking sh_t, @kamronbatman! You want to put this PR on hold until I spend my free time to develop an exploit for your auth bypass vulnerability?
> 
> We could discuss this vulnerability, I could explain it to you - but you demand that I write an exploit for you?
> 
> Hey man, I'm trying to help you with your project. I already spent a considerable amount of time on a piece of shitty code that I've never used and will never use. I left UO some 25 years ago. I just love to code. But I don't like this sh*t. Enough!

We have a saying in America. If you are going to take the shot, you better not miss. You missed buddy. And you missed by far. Good luck to you with other projects.

Projects/Server/Network/NetState/NetState.cs
MaxKellermann: @Bohicatv,

> Because, If the server is checking packets too slowly or missing them because the handler is busy, it could lead to a delay in processing player inputs or server responses. This would certainly result in a piling cascade of desync, even on modern hardware, where the player's actions are not reflected in real-time.

I think you misunderstand this PR. It only delays the one client whose packet handler is too slow.
- without this PR, the whole shard freezes (all clients)
- with this PR, only one client freezes

This is why I say this PR is a strict improvement: it turns a huge problem into a small problem.

Of course, packet handlers should be as fast as possible, and delaying one client for very long is bad. But this is not what this PR is about. It does not add any slow packet handlers or makes existing ones slower - in fact it changes no packet handler at all. It just prepares the infrastructure for changing a slow packet handler to asynchronous mode.

I did this (make slow packet handler asynchronous) for a ModernUO fork of a certain shard and it solved a grave problem. I submitted the basic patch here, too, because I'm convinced that generic improvements should be shared with everybody - upstream first! Unfortunately something went wrong here.

> To go even further, this would effectively allow duping without issues as it could be measured as the `_packetHandlerTask` could send multiple packets to account for the lost ones.

I don't get it at all. Why should any asynchronous packet handler sent multiple packets for the lost ones? How do packets get lost? And how does the handler know that packets were lost? And how does this allow duping?

PR #2035, opened 2024-12-30
TcpServer: simplify IPEndPoint logging

MaxKellermann:
The default IPEndPoint formatter does exactly this, and we don't need to reimplement that.


MaxKellermann:
Note: I won't sign the CLA, see https://github.com/modernuo/ModernUO/pull/2034#issuecomment-2565512972
If that means I can't contribute, then be it.

kamronbatman:
Using string interpolation breaks serilog's structure logging. It also breaks the async/threeded nature of the log pipeline by pushing the interpolation synchronously.

PR #2036, opened 2024-12-30
TcpServer: show address and exception message in log

MaxKellermann:


MaxKellermann:
Note: I won't sign the CLA, see https://github.com/modernuo/ModernUO/pull/2034#issuecomment-2565512972
If that means I can't contribute, then be it.

kamronbatman:
Same as the other PR, let's do it according to Serilog's recommendations.

Since you will not sign CLA, I'll make the proper changes when I get time.

PR #2040, opened 2024-12-31
feat: basic Docker support

IcculusC:


Dockerfile
```diff
@@ -0,0 +1,44 @@
+# syntax=docker.io/docker/dockerfile:1.7-labs
+ARG DOTNET_VERSION=9.0.101
+ARG OS=linux
+ARG ARCH=x64
+
+FROM ubuntu:24.10 AS dependencies
```
IcculusC, line 6: arbitrary base image

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ARG OS
+
+RUN apt-get update -y \
+    && apt-get install -y libicu-dev curl
+
+RUN curl -L https://dotnet.microsoft.com/download/dotnet/scripts/v1/dotnet-install.sh -o dotnet-install.sh && \
+    chmod +x dotnet-install.sh && \
+    ./dotnet-install.sh --version ${DOTNET_VERSION} --architecture ${ARCH} --os ${OS}
+
+ENV PATH $PATH:/root/.dotnet
```
IcculusC, line 18: we have to update the `PATH` here so the `publish` stage can run the `dotnet` commands it needs to run

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+COPY . .
+
+RUN ./publish.sh release ${OS} ${ARCH}
+
+FROM dependencies AS deploy
+ARG DOTNET_VERSION
+ARG ARCH
+
+ENV DOTNET_ROOT=/root/.dotnet
```
IcculusC, line 34: updating the `PATH` in the `dependencies` stage above doesn't enable the `ModernUO` binary to find that installation

instead, we have to set `DOTNET_ROOT` before we run the binary

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+        ARCH: "${ARCH:-amd64}"
+    environment:
+      UO_DIRECTORY: ${UO_DIRECTORY:-/bin/UO}
+    ports:
+      - "2593:2593"
+      - "12000:12000"
+    volumes:
+      - "./Distribution/Saves:/bin/modernuo/Saves"
+      - "./Distribution/Configuration:/bin/modernuo/Configuration"
+      - "$UO_DIRECTORY:$UO_DIRECTORY"
```
IcculusC, line 16: tl;dr my UO directory is something like `/Users/icculusc/Downloads/UO`, so that's what it is in the `modernuo.json` from running it on the metal, this maps it directly to that directory, this is a bad solution to be fair

.github/workflows/create-release.yml
```diff
@@ -37,3 +39,10 @@ jobs:
         prerelease: false
         generate_release_notes: true
         token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+
+  build_and_push_image:
+    needs: [release]
+    uses: ./.github/workflows/publish_docker_image.yml
+    with:
+      version: ${{ needs.release.outputs.version }}
+    secrets: inherit
```
IcculusC, line 48: Rider really doesn't like this, but I'm pretty sure it's correct

PR #2042, opened 2024-12-31
feat: Add IP geolocation support by loading CIDR ranges and identifying countries for IP addresses

Bohicatv:


kamronbatman:
This looks great. Note: firewalling should be done on the edge/proxy.

kamronbatman:
I think we have utility functions to work with this already. I can clean this up later. However I am not sure if I want to commit a list only IPs for whole countries, even given the major DDoS issue.

IcculusC:
> However I am not sure if I want to commit a list only IPs for whole countries, even given the major DDoS issue.

to this point, feels like we'd be better served loading this by some means at runtime (once on application launch)

Bohicatv:
This features is just for the logger and any actual firewalling should be done as you see fit, but this should serve only for information purposes. A mere stepping stone of the technology. I plan to author a major overhaul of how MUO sees incoming connections and their origins based on a comprehensive list of an array of problematic countries and VPN nodes, but the code base is very bad in trying to accomplish this so don't hold your breath.

kamronbatman:
> This features is just for the logger and any actual firewalling should be done as you see fit, but this should serve only for information purposes. A mere stepping stone of the technology. I plan to author a major overhaul of how MUO sees incoming connections and their origins based on a comprehensive list of an array of problematic countries and VPN nodes, but the code base is very bad in trying to accomplish this so don't hold your breath.

I am definitely in favor of cleaning up the networking / TCPServer / netstate code.
Regardless though, I don't think the game server itself would do a good job of thwarting large (we have seen 600Gbps+ and 10mill/min) DDoSes.

Bohicatv:
I agree. My idea is to develop tools and features that create a comprehensive information hub within MUO, as I believe this is something we currently lack. If we treat MUO as a layer in the overall architecture for DDoS mitigation, it would be valuable to gather and analyze more detailed information about an attack.

PR #2043, opened 2024-12-31
added realAddress feature for container NAT networking

gshearer:
Added a runtime user configurable option "realAddress" -- if configured, MUO will tell the client to connect to this configured address if the client is coming from an IsPrivateNetwork() address.

This allows for container scenarios that are NAT'd as opposed to bridged, so that both local and outside clients can play simultaneously.


Projects/UOContent/Misc/ServerList.cs
```diff
@@ -70,6 +80,12 @@ public static void Initialize()
                     logger.Information("Server listing address set from config: {address}", _publicAddress);
                 }
             }
+
+            if(realAddress != null)
```
kamronbatman, line 84:
```suggestion
            if (realAddress != null)
```

Projects/UOContent/Misc/ServerList.cs
```diff
@@ -92,16 +110,25 @@ public static void OnServerListEvent(GatewayServer.ServerListEventArgs e)
...
                         localAddress = _publicAddress;
                     }
                 }
 
+                // if a realAddress is configured AND the client is connecting
+                // from an IsPrivateNetwork(), hand out the user-configured realAddress
+                // instead of the one we publish for the server listing
+                // this allows LAN-local clients to connect directly to the server
+                // rather than going to their default-gateway which probably won't work
+                if(IsPrivateNetwork(ripep.Address) && _realAddress != null)
```
kamronbatman, line 127:
```suggestion
                if (IsPrivateNetwork(ripep.Address) && _realAddress != null)
```

kamronbatman:
Thank you for submitting this. I believe this will work, but I am wondering if we should instead go another direction.

Currently we have all of the listeners, which by default would be all interfaces. I am wondering if it makes sense to get the IP range for each of those addresses, and if the connecting user is from one of those ranges, serve the IP from that listener. So for each listener, if the IP is within one of the private ranges, then we use that, otherwise if it is public, we serve the public address.

If someone containerizes or NATs using non-private networks, then we may want to look into amending/adding onto the network range using a config file (gateway-ip-ranges.json or whatever, I am making it up).

Let me know your thoughts on this approach since I believe it would algorithmically solve the issue without the need for an explicit prompt?

Projects/Server/Configuration/ServerConfigurationPrompts.cs
```diff
@@ -118,6 +118,54 @@ internal static List<string> GetDataDirectories()
...
+            ipStr = Console.ReadLine();
+
+            IPEndPoint ip;
+            if (string.IsNullOrWhiteSpace(ipStr))
+            {
+                ipStr = null;
+                break;
+            }
+
+            else
```
kamronbatman, line 146:
```suggestion
            }
            else
```

Projects/UOContent/Misc/ServerList.cs
```diff
@@ -70,6 +80,12 @@ public static void Initialize()
                     logger.Information("Server listing address set from config: {address}", _publicAddress);
                 }
             }
+
+            if (realAddress != null)
+            {
+                Resolve(realAddress, out _realAddress);
+                logger.Information("Local clients will be told to connect to realAddress: {address}", _realAddress);
```
kamronbatman, line 87:
```suggestion
                logger.Information("Local clients will be told to connect to realAddress: {Address}", _realAddress);
```

PR #2048, opened 2025-01-02
fix: Add champion teleporter configuration for Trammel

Bohicatv:


kamronbatman (approved):

kamronbatman:
Per my comment in discord. @Bohicatv - Do you mind taking a look at the existing file for Felucca and seeing if any of those need to be copied over too?

Thank you!

PR #2051, opened 2025-01-03
fix: Fixes Lich Lord minions from attacking their master and each other

Bohicatv:


kamronbatman:
I think I fixed this already recently. Can you double check before we decide on this PR?

Bohicatv:
I got reports that it was still bugged yesterday afternoon. I tested and was able to reproduce it yesterday. This PR will fix it in its entirety.

kamronbatman:
> I got reports that it was still bugged yesterday afternoon. I tested and was able to reproduce it yesterday. This PR will fix it in its entirety.

It doesn't fix the problem generally though. So it might be better to fix IsEnemy.

Bohicatv:
You're right, but I thought both PRs were needed since you did an attempt to fix it recently.

PR #2054, opened 2025-01-05
feat: events for Help system

IcculusC:
It would be much simpler to expose pages to external webhooks such as Discord or Slack without significantly modifying the core codebase if we adopt some simple `event`s for actions taken by players and staff.  Given this addition an ambitious ModernUO developer (probably me) could [create an addon](https://github.com/modernuo/ModernUO/tree/addon-tutorial/Projects/AddonExample) that implements the aforementioned webhook behavior by adding delegates to the events in this PR.

This PR is a suggested approach, it includes the following events:
* `PageEnqueued`
* `PageRemoved`
* `PageHandlerChanged`
* `PageWaiting`
* `StuckMenu`


kamronbatman:
I agree, and I assume the external system would be it's own project in this case.

PR #2058, opened 2025-01-09
feat: Add shutdown button with 15-minute delay and notification messages

Bohicatv:
Screenshot 2025-01-09 031708

### DIFF
--- a/Projects/UOContent/Gumps/AdminGump.cs
+++ b/Projects/UOContent/Gumps/AdminGump.cs
@@ -2152,23 +2152,29 @@ namespace Server.Gumps
                                 }
                             case 205:
                                 {
-                                    sender.Mobile.SendMessage("The shard will shutdown in 15 minutes for maintenance.");
-                                    Timer.DelayCall(TimeSpan.FromMinutes(15), () =>
+                                    int[] delays = { 15, 10, 5, 4, 1 };
+                                    string[] messages = 
                                     {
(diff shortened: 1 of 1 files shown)


kamronbatman:
Would it make sense to add a small delay text box to enter in the minutes?

Bohicatv:
Imo, 15 minutes is plenty of time for players to finish up tasks and to get to a safe place before the server goes down. I don't think its necessary.

Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -2147,6 +2150,27 @@ public override void OnResponse(NetState sender, in RelayInfo info)
                                     Shutdown(true, false);
                                     break;
                                 }
+                            case 205:
+                                {
+                                    sender.Mobile.SendMessage("The shard will shutdown in 15 minutes for maintenance.");
+                                    Timer.DelayCall(TimeSpan.FromMinutes(15), () =>
```
kamronbatman, line 2156: I am sorry to be nitpicky. Can we do a single timer delay and either call the next timerdelay recursively, or make it an interval with a count (4 or 5, whatever is needed accordingly) and set the Interval in the tick callback. So first one would be delay of 10 minutes and interval of 5 minutes, then after the second tick set interval to 4 minutes, set interval to 1 minute, then shutdown.

I know this is probably confusing, let me know if you need examples, or if this doesn't make sense. Also I am open to a better, easier to read approach.

kamronbatman:
Thank you Bohica!!

PR #2060, opened 2025-01-12
fix: Refactor movement logic in BaseAI to prevent unnecessary direction changes

Bohicatv:


Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2132,7 +2132,11 @@ public virtual MoveResult DoMoveImpl(Direction d, bool badStateOk)
 
         // This makes them always move one step, never any direction changes
         // TODO: This is firing off deltas which aren't needed. Look into replacing/removing this
-        m_Mobile.Direction = d;
+        if (m_Mobile.Direction != d)
```
kamronbatman, line 2135: Do we need to clip off run (mask check?)

Bohicatv:
In theory, yes. Still trying to figure out if the bits would actually be ignored. I will commit what I was going to update it to and you can tell me if its logical, but this movement logic is a pain as you know.

kamronbatman:
> In theory, yes. Still trying to figure out if the bits would actually be ignored. I will commit what I was going to update it to and you can tell me if its logical, but this movement logic is a pain as you know.

I'm trying to remember if I made this change but ended up reverting, or not merging or something. I'll check it out when I get back.

Bohicatv:
Need more time to work through all of BaseAI.

PR #2061, opened 2025-01-12
feat: Update spawner generation paths to include all .json in subdirectories for all maps

Bohicatv:


IcculusC:
would be nice to have a gump for this so it's easier to spawn what you need, but also make sure you get everything you need

kamronbatman:
Thank you!

PR #2063, opened 2025-01-13
fix: Fixes AnimalLoreGump to correctly display "None" if no previous condition is met

Bohicatv:


kamronbatman:
I don't think this is a functional change, but I appreciate that it might make debugging easier.

PR #2065, opened 2025-01-14
fix: Fix FillableContainer default WeaponSmith

Reetus:
Not sure if there is a way to set a default value when using SerializableProperty...

But FillableContentType.None is -1, whilst FillableContentType.Weaponsmith is 0, and AcquireContent doesn't do anything if the ContentType is not .None, therefore all FillableContainer's end up being Weaponsmith

### DIFF
--- a/Projects/UOContent/Items/Containers/Fillable Containers/FillableContainer.cs
+++ b/Projects/UOContent/Items/Containers/Fillable Containers/FillableContainer.cs
@@ -22,7 +22,7 @@ public abstract partial class FillableContainer : LockableContainer
     public FillableContainer(int itemID) : base(itemID)
     {
         Movable = false;
-        ContentType = FillableContentType.None;
+        _contentType = FillableContentType.None;
     }
 
     public virtual int MinRespawnMinutes => 60;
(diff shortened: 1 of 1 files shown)


kamronbatman (dismissed):

Projects/UOContent/Items/Containers/Fillable Containers/FillableContainer.cs
```diff
@@ -19,7 +19,11 @@ private void DeserializeRespawnTimer(TimeSpan delay)
         }
     }
 
-    public FillableContainer(int itemID) : base(itemID) => Movable = false;
+    public FillableContainer(int itemID) : base(itemID)
+    {
+        Movable = false;
+        ContentType = FillableContentType.None;
```
kamronbatman, line 25: I approved but I may change this to the field instead of the property to avoid the overhead setter.

Bohicatv:
I resolved this awhile ago and all I did was change each container type instead of assigning a specific type from the list.
You can do something like 

```
        if (ContentType != FillableContentType.Weaponsmith)
        {
            return;
        }

        ContentType = FillableContentType.Library;
```

Projects/UOContent/Items/Containers/Fillable Containers/FillableContainer.cs
kamronbatman, line 25:
```suggestion
        _contentType = FillableContentType.None;
```

PR #2068, opened 2025-01-15
fix: Increases rangePerception in BaseMount constructor from 10 to 16

Bohicatv:
This should not be 10. It is 16 on OSI from what i can tell.


kamronbatman:
The perception is for pets, or wild?

Bohicatv:
When you tame, they don't take on a different class. They keep Basemount. Basemount inherits basecreature, so can remove or change it to 16.

Bohicatv:
good idea i guess to keep both for customization sake and to prevent accidental overrides.

PR #2079, opened 2025-01-19
ref: BaseAI, Movement, Speech (Active Dev PR, WIP)

Bohicatv:
Refactor of BaseAI.
This iteration is nearing merger.
Usage/Testing is needed to further find potential bugs and non-OSI behaviors from the community.
Please message me by using this link below if you encounter something worth noting. Thanks.
https://discord.com/channels/751317910504603701/1329631466690318426


Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -25,6 +48,8 @@ public enum AIType
     AI_Healer,
     AI_Vendor,
     AI_Mage,
+    AI_Mage_Fire, // REMOVE ME BEFORE PR
+    AI_Mage_Energy, // REMOVE ME BEFORE PR
```
IcculusC, line 52: did you forget 😂

Bohicatv: Yes, I did, thank you. Just some custom types I currently have.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -144,172 +146,212 @@ public ActionType Action
...
+
+    private static bool IsFriendlyOrder(OrderType order) =>
+        order is OrderType.Follow or OrderType.Stay or OrderType.Stop;
+
+    private bool CanAttackTarget(Mobile from, Mobile target)
+    {
+        if (target is BaseCreature creature && creature.IsScaryToPets && m_Mobile.IsScaredOfScaryThings)
         {
-            return;
+            m_Mobile.SayTo(from, "Your pet refuses to attack this creature!");
```
kamronbatman, line 327: Use `1080051` -> `Your pet refuses to attack this creature!` - Not sure what version it was added in, sometime after UOML 5.0.9.1.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1834,242 +1760,274 @@ public virtual bool DoOrderTransfer()
...
+                SendTransferRefusalMessages(from, to, 1043250, 1043251);
+                // 1043250: The pet refuses to be transferred because it will not obey you sufficiently.~3_BLANK~
+                // 1043251: The pet will not accept you as a master because it does not trust ~2_NAME~.~3_BLANK~
             }
-            else if (TransferItem.IsInCombat(m_Mobile))
+            else if (IsInCombatState())
             {
-                from.SendMessage("You may not transfer a pet that has recently been in combat.");
-                to.SendMessage("The pet may not be transferred to you because it has recently been in combat.");
+                from.SendLocalizedMessage(1042561);
```
kamronbatman, line 1788: 1042561 is `Please dismount first.`

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1834,242 +1760,274 @@ public virtual bool DoOrderTransfer()
...
+                // 1042561: You may not transfer a pet that has recently been in combat.
+                to.SendLocalizedMessage(1042562);
```
kamronbatman, line 1790: `1042562` is `You do not own that pet!`

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2820,409 +2795,487 @@ public virtual void DetectHidden()
...
+                !ValidateFollowerLimit(to) || 
+                IsInCombat(m_Creature))
             {
                 return false;
             }
 
+            return true;
+        }
+    
+        private bool ValidateYoungStatus(Mobile from, Mobile to)
```
kamronbatman, line 3096: Duplicate method logic with `IsYoungPlayerMismatch`?

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2820,409 +2795,487 @@ public virtual void DetectHidden()
...
                 Stop();
                 return;
             }
 
-            if (m_Owner.m_Mobile.Map == null || m_Owner.m_Mobile.Map == Map.Internal)
+            // update interval based on current state and speed
+            Interval = TimeSpan.FromMilliseconds(GetBaseInterval(m_Owner));
+            m_Owner.m_Mobile.OnThink();
+
+            if (ShouldStop())
```
kamronbatman, line 3237: I never liked the idea that `OnThink()` could result in the mob dying/getting deleted. I wish there was a cleaner way?

Bohicatv: https://github.com/modernuo/ModernUO/pull/2079/commits/a3f07227ebb92bb2e6d7649d371e85350bc580e6: Instead of deletion, the mobile is hidden and a GM can search for them.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -144,172 +146,212 @@ public ActionType Action
...
         var isFriend = !isOwner && m_Mobile.IsPetFriend(from);
 
         if (!isOwner && !isFriend)
         {
-            return;
+            return false;
         }
 
-        if (isFriend && order != OrderType.Follow && order != OrderType.Stay && order != OrderType.Stop)
+        return !isFriend || IsFriendlyOrder(order);
```
kamronbatman, line 317: Could this block be reduced to?
```cs
return isOwner || isFriend || !isFriendlyOrder(order);
```

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,771 +370,604 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
+            // 501482: I am too busy fighting to deal with thee!
+            return;
+        }
+    
+        var actions = new Dictionary<int, Action>
+        {
+            { 0x9D, () => HandleMove() }, // *move*
+            { 0x9E, () => HandleTime() }, // *time*
+            { 0x6C, () => HandleTraining(e.Mobile) } // *train*
+        };
```
kamronbatman, line 390: Note: Move this to a static variable.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -144,172 +146,212 @@ public ActionType Action
...
         }
     }
 
     public long NextMove { get; set; }
 
+    // OSI, DetectHidden >= 50.0?
     public virtual bool CanDetectHidden => m_Mobile.Skills.DetectHidden.Value > 0;
-
+    
     public virtual bool WasNamed(string speech)
```
kamronbatman, line 162: We should assume `WasNamed` is idempotent against the same speech. This means all of the uses for each switch/loop/sub-function will return the same value. We can refactor by reducing the number of times we call WasNamed to once and use that bool value where we need it.

For loops where all commands require WasNamed to be true, we can do a single check before the loop too.

Projects/UOContent/Mobiles/AI/BaseAI.cs
kamronbatman: I am not sure how I feel about hiding the mob. I'll think about it.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1276,44 +1075,56 @@ public virtual bool DoOrderCome()
...
+    private void RefreshCombatantState()
+    {
+        if (IsValidCombatant(m_Mobile.Combatant))
+        {
+            m_Mobile.Warmode = true;
+        }
+        else
+        {
+            m_Mobile.Warmode = false;
+        }
```
kamronbatman, line 1127: Can be compressed to this:

```suggestion
        m_Mobile.Warmode = IsValidCombatant(m_Mobile.Combatant);
```

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1834,242 +1711,274 @@ public virtual bool DoOrderTransfer()
...
+    }
 
-        return true;
+    private bool IsInCombatState()
+    {
+        return m_Mobile.Combatant != null || m_Mobile.Aggressors.Count > 0 ||
+        m_Mobile.Aggressed.Count > 0 || Core.TickCount < m_Mobile.NextCombatTime;
+    }
+    
+    private bool IsValidTransferRequest(Mobile from, Mobile to)
```
kamronbatman, line 1767: I see duplicate validity check logic, we should consolidate these functions to one?

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2550,267 +2497,248 @@ public virtual bool AcquireFocusMob(int iRange, FightMode acqType, bool bPlayerO
...
+        }
 
-    private bool IsHostile(Mobile from)
+        if (m_Mobile.Summoned && m_Mobile.SummonMaster != null)
+        {
+            return m == m_Mobile.SummonMaster || 
+                   !SpellHelper.ValidIndirectTarget(m_Mobile.SummonMaster, m) ||
+                   (pm != null && m_Mobile.IsAnimatedDead) || 
+                   (m_Mobile.IsAnimatedDead && bc?.IsAnimatedDead == true) ||
+                   (m_Mobile.IsAnimatedDead && bc?.Controlled == true);
```
kamronbatman, line 2670:
```suggestion
                   (m_Mobile.IsAnimatedDead && (bc?.IsAnimatedDead == true) || bc?.Controlled == true);
```

Projects/UOContent/Mobiles/AI/BaseAI.cs
Bohicatv: one is for friending a pet and one is for transferring transferring a pet

Projects/UOContent/Mobiles/AI/BaseAI.cs
Bohicatv: must have been added after ML

Projects/UOContent/Mobiles/AI/BaseAI.cs
kamronbatman: Sorry, that is not what I meant. You changed the message from text to a localization number, but that isn't the correct number. In fact, I don't see that message localized.

Projects/UOContent/Mobiles/AI/BaseAI.cs
kamronbatman: Same as above.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,596 +365,450 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
-                {
-                    // Excuse me?
-                    m_Mobile.PublicOverheadMessage(MessageType.Regular, 0x3B2, 501516);
-                    WalkRandomInHome(2, 2, 1);
-                }
+                m_Mobile.PublicOverheadMessage(MessageType.Regular, 0x3B2, 501516);
+                // 501516: Excuse me?
+                m_Mobile.Location = new Point3D(m_Mobile.Location.X + 
+                Utility.RandomMinMax(-1, 1), m_Mobile.Location.Y + 
+                Utility.RandomMinMax(-1, 1), m_Mobile.Location.Z);
```
kamronbatman, line 377: Why are we force moving them, potentially into a wall or obstruction, before we make them walk?

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,596 +365,450 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
+                WalkRandomInHome(4, 3, 1);
+                return;
```
kamronbatman, line 379: We can't return early, cause we are cutting off Line 394 through 404.
This is the same for L385 and L390.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,596 +365,450 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
-                        m_Mobile.Say(501505); // Alas, I cannot teach thee anything.
-                    }
-                }
+    public virtual void AllOnSpeechPet(SpeechEventArgs e)
+    {
+        if (e.Mobile.InRange(m_Mobile.Location, 14))
+        {
+            if (e.HasKeyword(0x164)) // all come
+            {
+                HandleComeCommand(e.Mobile, true);
```
kamronbatman, line 413: isOwner checks are missing, assuming true all the time.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,596 +365,450 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
+            HandleGMCommands(e);
+        }
+    }
 
-                    if (!foundSomething)
-                    {
-                        m_Mobile.Say(501505); // Alas, I cannot teach thee anything.
-                    }
-                }
+    public virtual void AllOnSpeechPet(SpeechEventArgs e)
```
kamronbatman, line 407: We lost the ability to process multiple keywords at once.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,596 +365,450 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
+    {
+        if (e.Mobile.InRange(m_Mobile.Location, 14))
+        {
+            if (e.HasKeyword(0x164)) // all come
```
kamronbatman, line 411: The HasKeyword actually does a for-loop, so now we are doing a loop per keyword which isn't ideal. I know the loop/switch is ugly, but that might be a better option.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,596 +365,450 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
-            default:
-                {
-                    return false;
-                }
+        if (CheckFlee())
+        {
+            return true;
         }
+    
+        var actionHandlers = new Dictionary<ActionType, Func<bool>>
```
kamronbatman, line 688: Let's not make a dictionary hot. I don't think we gain anything over a switch statement.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -328,596 +365,450 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
+            { ActionType.Backoff, () => { m_Mobile.OnActionBackoff(); return DoActionBackoff(); } }
+        };
+    
+        return actionHandlers.TryGetValue(Action, out var handler) && handler();
     }
 
     public virtual void OnActionChanged()
     {
-        switch (Action)
+        var actionHandlers = new Dictionary<ActionType, Action>
```
kamronbatman, line 703: Same as above.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -993,106 +899,131 @@ public virtual bool DoActionCombat()
...
-
+    
         m_Mobile.Direction = m_Mobile.GetDirectionTo(combatant);
+    
         if (m_Mobile.TriggerAbility(MonsterAbilityTrigger.CombatAction, combatant))
         {
             if (m_Mobile.Debug)
             {
-                m_Mobile.DebugSay($"I used my abilities on {combatant.Name}!");
+                m_Mobile.DebugSay(string.Format("I used my abilities on: {0}", combatant.Name));
```
kamronbatman, line 930: Revert

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -993,106 +899,131 @@ public virtual bool DoActionCombat()
...
         else
         {
             if (m_Mobile.Debug)
             {
-                m_Mobile.DebugSay("I stopped being on guard");
+                m_Mobile.DebugSay("I stopped being on guard.");
             }
 
-            Action = ActionType.Wander;
+            if (Action != ActionType.Wander)
```
kamronbatman, line 961: We don't need any of these action should not already equal what we want checks cause it is done in the setter.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -993,106 +899,131 @@ public virtual bool DoActionCombat()
...
+    }
 
-    public virtual bool DoActionInteract() => true;
+    private void DebugSay(string message)
+    {
+        if (m_Mobile.Debug)
+        {
+            m_Mobile.DebugSay(message);
+        }
+    }
```
kamronbatman, line 1021: We need a better way of doing this, and this makes it worse because we don't want to do the interpolation into a string unless debug is enabled. I'll look into a better way.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -993,106 +899,131 @@ public virtual bool DoActionCombat()
...
+            if (Action != ActionType.Guard)
+            {
+                Action = ActionType.Guard;
+            }
             return true;
         }
-
-        if (WalkMobileRange(from, 1, true, m_Mobile.RangePerception * 2, m_Mobile.RangePerception * 3))
+    
+        if (WalkMobileRange(from, 1, true, m_Mobile.RangePerception * 2, m_Mobile.RangePerception * 1))
```
kamronbatman, line 988: Maximum is lower than minimum here. Let's change this to RangePerception -> RangePerception * 2

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1103,160 +1034,182 @@ public virtual bool Obey() =>
...
-            case OrderType.Stay:
-                {
-                    m_Mobile.ControlMaster.RevealingAction();
-                    m_Mobile.SetCurrentSpeedToPassive();
-                    m_Mobile.PlaySound(m_Mobile.GetIdleSound());
-                    m_Mobile.Warmode = false;
-                    m_Mobile.Combatant = null;
-                    break;
-                }
+        m_Mobile.Home = m_Mobile.ControlMaster.Location;
```
kamronbatman, line 1130: This is a functionality change. We are not setting `Home` for `HandlePassiveOrder()`, for stop/none, but we do set it for `Guard`? Is this on purpose?

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1103,160 +1034,182 @@ public virtual bool Obey() =>
...
+        {
+            m_Mobile.ControlOrder = OrderType.None;
+        }
 
-            case OrderType.Follow:
-                {
-                    m_Mobile.ControlMaster.RevealingAction();
-                    m_Mobile.SetCurrentSpeedToActive();
-                    m_Mobile.PlaySound(m_Mobile.GetIdleSound());
+        m_Mobile.PlaySound(m_Mobile.GetAttackSound());
```
kamronbatman, line 1137: I don't think we want an attack sound here? Wouldn't we want the idle sound?

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -1103,160 +1034,182 @@ public virtual bool Obey() =>
...
-                }
+        if (!m_Mobile.Warmode)
+        {
+            m_Mobile.Warmode = true;
+        }
+        if (m_Mobile.FocusMob != null)
+        {
+            m_Mobile.FocusMob = null;
+        }
+        if (m_Mobile.CurrentSpeed != m_Mobile.PassiveSpeed)
```
kamronbatman, line 1125: All of these checks are handled for us in the setter. We can get rid of them for all of these properties.

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -363,30 +363,64 @@ public virtual bool HandlesOnSpeech(Mobile from)
...
         {
             if (e.HasKeyword(0x9D)) // *move*
             {
-                m_Mobile.PublicOverheadMessage(MessageType.Regular, 0x3B2, 501516);
-                // 501516: Excuse me?
-                m_Mobile.Location = new Point3D(m_Mobile.Location.X + 
-                Utility.RandomMinMax(-1, 1), m_Mobile.Location.Y + 
-                Utility.RandomMinMax(-1, 1), m_Mobile.Location.Z);
-                WalkRandomInHome(4, 3, 1);
+                if ((DateTime.Now - m_lastOrder).TotalSeconds < 5)
```
kamronbatman, line 375: FYI: Use `Core.Now` instead of `DateTime.Now`

Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -363,30 +363,64 @@ public virtual bool HandlesOnSpeech(Mobile from)
         return from.Alive && from.InRange(m_Mobile.Location, 3) && m_Mobile.IsHumanInTown();
     }
 
+    private DateTime m_lastOrder = DateTime.MinValue;
```
kamronbatman, line 366: We should get away from the legacy `m_` variables and use `_camelCase`.

Projects/UOContent/Mobiles/AI/BaseAI.cs
Bohicatv: The original guard had a timer `m_NextStopGuard = Core.TickCount + (int)TimeSpan.FromSeconds(10).TotalMilliseconds;` which I got rid of. Home is set to controlmaster, because its guarding that target. If we set a new home for passive or none, it may mess with home locations on spawners.

Projects/UOContent/Mobiles/AI/BaseAI.cs
Bohicatv: I'm not sure, functionality wise, how I can use switches here with out needing to list the `e.HasKeyword` anyway.

PR #2083, opened 2025-01-21
fix: Enabled vendor sell SayTo with correct code.

Senion31:


Projects/UOContent/Mobiles/Vendors/BaseVendor.cs
```diff
@@ -560,8 +560,9 @@ public virtual bool OnSellItems(Mobile seller, List<SellItemResponse> list)
                     }
                 }
             }
-            // no cliloc for this?
-            // SayTo( seller, true, "Thank you! I bought {0} item{1}. Here is your {2}gp.", Sold, (Sold > 1 ? "s" : ""), GiveGold );
+            
+            // No CliLoc available for this message; enabled string interpolation for readability.
+            SayTo( seller, true, $"Thank you! I bought {Sold} item{(Sold > 1 ? "s" : "")}. Here is your {GiveGold}gp.");
```
kamronbatman, line 565: If I remember correctly, this was removed from OSI sometime well before AOS. Would somebody be able to verify this?

kamronbatman:
After some discussion, let's skip on adding this back in, especially since we don't know when (if?) it was used.

PR #2084, opened 2025-01-22
fix: Fixes rangePerception to 16 to match OSI

Bohicatv:


kamronbatman:
To clarify, is this for command perception or acquire target (for combat)?

I think the range for combat is era specific. The range for commands used to be 10 or 12 unless you use yell and then it's 16, and that was changed at some point to not require yell.

Bohicatv:
All of the commands and the tile ranges are all handled inside of BaseAI along with the context menus, etc. rangePerception effects the mobile's ability to register other mobiles as a target. So, if the mobile is a mount and its range is capped at 10, it won't follow you unless you are within 10 tiles which is not OSI.

kamronbatman:
I think range perception (at this point) might be too broad. I am wondering if we need to split it into two properties because I believe it is also used for acquiring targets to fight. Can you confirm?

Bohicatv:
I don't understand why that is an issue.

kamronbatman (approved):

PR #2090, opened 2025-01-26
fix: Aquarium no longer uses list to kill fish

kamronbatman:


Projects/UOContent/Items/Aquarium/Aquarium.cs
```diff
@@ -551,25 +551,21 @@ public int FoodNumber() =>
...
-                        toKill.Add(fish);
-                    }
+                    toKill.Add(fish);
                 }
             }
 
-            while (amount > 0 && toKill.Count > 0)
+            toKill.Shuffle();
+
+            for (var i = 0; i < amount; i++)
```
kamronbatman, line 566: I need to fix this, it should not loop more than count number of times.

PR #2091, opened 2025-01-26
fix: Fixes door monster LOS exploit & AOS House Gump NPE

mark1145:
It's an old school RunUO bug that people squirreled away to day 1 balron/boss monsters. Essentially by standing on a door way, the HasLineOfSight() of sight check returns false and the monster just stands there, doing nothing, not even returning melee swings

### DIFF
--- a/Projects/UOContent/Gumps/Houses/HouseGumpAOS.cs
+++ b/Projects/UOContent/Gumps/Houses/HouseGumpAOS.cs
@@ -115,11 +115,14 @@ namespace Server.Gumps
             {
                 var lines = m_House.Sign.GetName().Wrap(10, 6);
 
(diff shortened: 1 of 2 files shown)


Projects/UOContent/Gumps/Houses/HouseGumpAOS.cs
```diff
@@ -115,11 +115,14 @@ public HouseGumpAOS(HouseGumpPageAOS page, Mobile from, BaseHouse house) : base(
             {
                 var lines = m_House.Sign.GetName().Wrap(10, 6);
 
-                for (int i = 0, y = (114 - lines.Count * 14) / 2; i < lines.Count; ++i, y += 14)
```
kamronbatman, line 118: This is a separate fix? What was the issue here?

PR #2102, opened 2025-02-01
fix: Ensures world save paths are validated

Bohicatv:


Projects/Server/Serialization/GenericEntityPersistence.cs
```diff
@@ -65,6 +65,13 @@ public GenericEntityPersistence(string name, int priority, Serial minSerial, Ser
 
     public override void WriteSnapshot(string savePath, HashSet<Type> typeSet)
     {
+        if (string.IsNullOrEmpty(savePath))
```
kamronbatman, line 68: This is impossible to be null or empty. Did you have an actual issue? If so can you give some details?

Projects/Server/Serialization/GenericEntityPersistence.cs
```diff
@@ -65,6 +65,13 @@ public GenericEntityPersistence(string name, int priority, Serial minSerial, Ser
...
+        {
+            throw new ArgumentNullException(nameof(savePath));
+        }
+
+        var path = Path.Combine(savePath, "snapshot");
```
kamronbatman, line 73: Is this variable used?

Projects/Server/Serialization/GenericEntityPersistence.cs
Bohicatv: Saves were trying to pass empty paths.

PR #2103, opened 2025-02-02
fix: Fixes cannot see that issue with Map LOS Refactor.

kamronbatman:


Projects/Server/Geometry/Point3DList.cs
```diff
@@ -9,7 +9,7 @@ public class Point3DList
 
     public Point3DList()
     {
-        m_List = new Point3D[8];
+        m_List = new Point3D[16];
```
kamronbatman, line 12: Bumped this to 16 because `eye` level is 14, and those checks use 15 point locations, so it is pretty common.

Projects/Server/Maps/Map.cs
```diff
@@ -973,25 +973,21 @@ public bool LineOfSight(Point3D origin, Point3D destination)
...
-        var path = new Point3DList();
-        TileFlag flags;
-
         if (origin == destination)
         {
             return true;
         }
 
-        if (path.Count > 0)
+        var end = destination;
```
kamronbatman, line 981: Crux of the change. put back `end` which is not potentially swapped.

PR #2106, opened 2025-02-03
fix: Fixes all summon spells so they use Summoned variation of monster

Bohicatv:


Projects/UOContent/Spells/Eighth/AirElemental.cs
```diff
@@ -49,14 +49,7 @@ public override void OnCast()
...
                 };
 
-                if (Core.AOS)
-                {
-                    SpellHelper.Summon(new SummonedAirElemental(), Caster, 0x217, duration, false, false);
-                }
-                else
-                {
-                    SpellHelper.Summon(new AirElemental(), Caster, 0x217, duration, false, false);
-                }
```
kamronbatman, line 59: I wonder exactly what the distinction was on OSI when AOS was introduced. I'll have to do some digging.

Bohicatv: Unless OSI had a different system, wild mobiles should not have dispel chance, control slots, etc. I can see Air Elementals and the like having a chance to dispel, but I don't remember if you could actually dispel them unless they were summoned. Pretty certain only summoned creatures could be dispelled in such a manner. Seems game breaking otherwise.

kamronbatman: I just realized it is cause of the stats change.

PR #2111, opened 2025-02-05
fix: Ensures Item constructors are not accessing the properties of null objects

Bohicatv:


Projects/UOContent/Items/TreasureChests/TreasureChestLevel3.cs
```diff
@@ -76,25 +76,35 @@ public TreasureChestLevel3() : base(0xE41)
...
+            armor.Durability = (ArmorDurabilityLevel)Utility.Random(3);
+            armor.Quality = ArmorQuality.Regular;
+            DropItem(armor);
+        }
+        
+        // Shield
+        var shield = Loot.RandomShield();
+        if (shield != null)
         {
-            var item = Loot.RandomArmorOrShieldOrWeapon();
```
kamronbatman, line 81: This should never return a null item. Do we have a bug somewhere or is this caution?

PR #2113, opened 2025-02-05
fix: Fixes AR/Durability/Tool Uses for Pre-UOR and LBR

Bohicatv:


Projects/UOContent/Items/Armor/BaseArmor.cs
```diff
@@ -857,49 +887,61 @@ public CraftAttributeInfo GetResourceAttrs() =>
 
         public int GetProtOffset()
         {
-            return _protectionLevel switch
+            if (Core.T2A)
```
kamronbatman, line 890: Should not have changed this. Damn.

PR #2134, opened 2025-02-27
fix: Fixes TCPServer accept async, makes Firewall/IP Limiter multithreaded

kamronbatman:
### Summary

* Updates socket accept so it runs on background threads.
* Updates initial socket processing so it runs on background threads. 
* Updates IP Limiter so it is multithreaded.
* Updates Firewall so it is multithreaded.
* Adds first packet peek to detect garbage data.


Projects/UOContent/Misc/AdminFirewall.cs
```diff
@@ -128,9 +128,12 @@ public static bool Add(IFirewallEntry entry, bool save = true)
     public static void Save()
     {
         using var op = new StreamWriter(firewallConfigPath);
```
kamronbatman, line 130: Move this into the Read function to avoid possible disposed access.

PR #2136, opened 2025-03-04
fix: Fixes edge cases where a leader resigns and becomes promoted in their new guild.

mark1145:
Because the guild they resigned from, would not calculate a new GM if all the members were a low rank, and so when the server next restarted, they would be loaded up as a Guild Leader in whatever clan they joined

### DIFF
--- a/Projects/UOContent/Misc/Guild.cs
+++ b/Projects/UOContent/Misc/Guild.cs
@@ -1562,6 +1562,13 @@ namespace Server.Guilds
                 winner = m_Leader;
             }
 
+            //There MUST be a new guild master.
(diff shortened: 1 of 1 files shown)


Projects/UOContent/Misc/Guild.cs
```diff
@@ -1562,7 +1554,33 @@ public void CalculateGuildmaster()
...
+                {
+                    winner = Members.RandomElement();
+                }
+            }
+
+            if (winner == null)
+            {
+                m_Leader = null;
+            }
+            else if (m_Leader != winner)
```
kamronbatman, line 1583: If no leader was chosen, pick one at random from anyone that got votes, and if nobody got votes, choose one at random from all members, otherwise the guild leader is null, resulting in disbanding.

PR #2154, opened 2025-04-12
fix: Optimizes GetString to eliminate allocations

kamronbatman:
### Summary

* Optimized GetString by eliminating the intermediate string allocation.

** Note **: Encoding.GetChars() is still really inefficient, especiall when strings are not aligned or have regular ascii/unicode characters. Thankfully we generally don't have to worry about these odd edge cases, but if they happen then GetChars can allocate hundreds of bytes.

This is a benchmark for just the related changes. NonSpecial are just ascii characters, while the other tests include control codes.
```cs
| Method                         | Mean     | Error   | StdDev  | Gen0   | Allocated |
|------------------------------- |---------:|--------:|--------:|-------:|----------:|
| GetString                      | 306.7 ns | 5.96 ns | 6.86 ns | 0.0124 |     200 B |
| GetStringNotSpecial            | 257.0 ns | 4.83 ns | 4.52 ns | 0.0114 |     184 B |
| GetStringSpanHelpers           | 166.4 ns | 3.26 ns | 3.48 ns |      - |         - |
| GetStringSpanHelpersNotSpecial | 137.3 ns | 1.57 ns | 1.22 ns |      - |         - |
```

### DIFF
--- a/Projects/Server/Text/TextEncoding.cs
+++ b/Projects/Server/Text/TextEncoding.cs
@@ -1,6 +1,6 @@
 /*************************************************************************
  * ModernUO                                                              *
- * Copyright 2019-2023 - ModernUO Development Team                       *
+ * Copyright 2019-2025 - ModernUO Development Team                       *
  * Email: hi@modernuo.com                                                *
  * File: TextEncoding.cs                                                 *
  *                                                                       *
@@ -16,6 +16,7 @@
 using System;
 using System.Runtime.CompilerServices;
 using System.Text;
+using Server.Buffers;
 
 namespace Server.Text;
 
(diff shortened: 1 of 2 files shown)


Projects/Server/Text/TextEncoding.cs
```diff
@@ -114,33 +115,55 @@ public static int GetByteLengthForEncoding(this Encoding encoding) =>
             _          => 1
         };
 
-    [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    private static bool IsSafeChar(ushort c) => c is >= 0x20 and < 0xFFFE;
-
     public static string GetString(ReadOnlySpan<byte> span, Encoding encoding, bool safeString = false)
     {
-        string s = encoding.GetString(span);
```
kamronbatman, line 122: Honestly this is the crux of the change.

Projects/Server/Text/TextEncoding.cs
```diff
@@ -114,33 +115,55 @@ public static int GetByteLengthForEncoding(this Encoding encoding) =>
...
-
         if (!safeString)
         {
-            return s;
+            return encoding.GetString(span);
         }
 
-        ReadOnlySpan<char> chars = s.AsSpan();
+        var charCount = encoding.GetMaxCharCount(span.Length);
```
kamronbatman, line 125: We use GetMaxCharCount instead of GetCharCount since the latter is considerably slower and sometimes allocates.

Projects/Server/Text/TextEncoding.cs
```diff
@@ -114,33 +115,55 @@ public static int GetByteLengthForEncoding(this Encoding encoding) =>
...
+        Span<char> chars = charCount <= 256
+            ? stackalloc char[charCount]
+            : rentedChars = STArrayPool<char>.Shared.Rent(charCount);
 
-        for (int i = 0, last = 0; i < chars.Length; i++)
+        try
         {
-            if (!IsSafeChar(chars[i]) && i == chars.Length - 1)
+            var length = encoding.GetChars(span, chars);
+            chars = chars[..length];
```
kamronbatman, line 135: This still allocates with weird strings. I am not going to spend the time to guard against malicious input that could cause 2-5x perf hit or the 300+ byte allocation. It's unfortunate, but maybe someone knows of a way to check/sanitize a buffer for real characters and effectively bail out early if the string is corrupt/nonsense.

PR #2157, opened 2025-04-14
fix: Fixes mutate speech character limit

kamronbatman:
### Summary

* Fixes the accidental limitation of dead character speech (OoOo) to 256 characters.
* Optimizes it by 1.5x

```cs
| Method                  | Mean      | Error    | StdDev   | Allocated |
|------------------------ |----------:|---------:|---------:|----------:|
| ManuaLoopMutation       | 147.01 ns | 1.298 ns | 1.084 ns |         - |
| SpanLoopMutation        |  92.03 ns | 0.583 ns | 0.487 ns |         - |
```

### DIFF
--- a/Projects/Server/Mobiles/Mobile.cs
+++ b/Projects/Server/Mobiles/Mobile.cs
@@ -29,6 +29,7 @@ using Server.Text;
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using Server.Buffers;
 using CalcMoves = Server.Movement.Movement;
 
 namespace Server;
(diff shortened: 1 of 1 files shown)


Projects/Server/Mobiles/Mobile.cs
```diff
@@ -5371,13 +5372,31 @@ public virtual bool MutateSpeech(List<Mobile> hears, ref string text, ref object
...
+        finally
+        {
+            if (rentedChars != null)
+            {
+                STArrayPool<char>.Shared.Return(rentedChars);
+            }
         }
 
-        text = sb.ToString();
+        text = new string(chars[..length]);
```
kamronbatman, line 5399: This needs to move into the try/finally or the slice will be collected.

PR #2159, opened 2025-04-14
feat: Adds an Item/Mobile memory leak detector. Fixes minor leak in doors.

kamronbatman:


Projects/UOContent/Engines/Quests/Uzeraan Turmoil/Items/SchmendrickApprenticeCorpse.cs
```diff
@@ -27,6 +27,8 @@ public SchmendrickApprenticeCorpse() : base(GetOwner(), GetHair(), GetFacialHair
 
         _lantern = new Lantern { Movable = false, Protected = true };
         _lantern.Ignite();
+
+        Owner = null;
```
kamronbatman, line 31: Corpse holds onto a reference to the Owner until it decays. We can get rid of it immediately after the constructor runs.

Projects/UOContent/Items/Construction/Doors/BaseDoor.cs
```diff
@@ -442,6 +449,12 @@ public InternalTimer(BaseDoor door) : base(TimeSpan.FromSeconds(20.0), TimeSpan.
 
         protected override void OnTick()
         {
+            if (_door.Deleted)
```
kamronbatman, line 452: If a door is deleted while open, this timer never finishes.

PR #2170, opened 2025-05-01
fix: Fixes BitArray serialization

kamronbatman:
### Summary

Fixes serialization/deserialization edge cases with BitArray. If you use BitArray, you will need to migrate.

### Migration

1. Change the type in the migration JSON file (if there is one) from `BitArray` to `byte[]` for all the versions you need to migrate.
2. Then in the `MigrateFrom`, use the following function to convert the field from a byte[] back to the BitArray.

```cs
public static BitArray MigrateByteArrayToBitArray(ReadOnlySpan<byte> data, int bitLength)
{
    // ReadEncodedInt
    var byteLength = 0;
    var shift = 0;
    var index = 0;
    byte b;
    do
    {
        b = data[index++];
        byteLength |= (b & 0x7F) << shift;
        shift += 7;
    }
    while (b >= 0x80);

    var buffer = ArrayPool<byte>.Shared.Rent(byteLength);
    try
    {
        data.CopyTo(buffer.AsSpan(index));
        return new BitArray(buffer) { Length = bitLength };
    }
    finally
    {
        ArrayPool<byte>.Shared.Return(buffer);
    }
}
```

### DIFF
--- a/Projects/Server/Serialization/IGenericReader.cs
+++ b/Projects/Server/Serialization/IGenericReader.cs
@@ -14,6 +14,7 @@
  *************************************************************************/
 
 using System;
+using System.Buffers;
 using System.Collections;
 using System.IO;
 using System.Net;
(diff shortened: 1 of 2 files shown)


kamronbatman:
Migration example for Regions in a Box (RegionController):

```json
{
      "name": "RestrictedSpells",
      "type": "byte[]",
      "rule": "ArrayMigrationRule",
      "ruleArguments": [
        "byte",
        "PrimitiveTypeMigrationRule",
        ""
      ]
    }
```

PR #2182, opened 2025-05-11
fix: Fixes craft gump non-cliloc label offsets.

mdodkins:
This has been bothering me for a while:

before

Been broken forever... I just corrected the offset:

after


kamronbatman:
Oh wow! Nice catch, thanks!

PR #2190, opened 2025-05-18
Fixes direction snapping

Bohicatv:


Projects/UOContent/Mobiles/AI/MageAI.cs
```diff
@@ -846,7 +846,9 @@ public override bool DoActionCombat()
             RunTo(c);
         }
 
-        if (m_Mobile.Spell != null || !m_Mobile.InRange(c, 1) || Core.TickCount - m_Mobile.LastMoveTime > 800)
+        if ((m_Mobile.Spell != null && m_Mobile.Spell.IsCasting) || m_Mobile.InRange(c, 1) 
+            || (m_Mobile.Spell == null && !m_Mobile.InRange(c, 1)
+            && Core.TickCount - m_Mobile.LastMoveTime > 800))
```
kamronbatman, line 851: If we treat `m_Mobile.Spell?.IsCasting == true` and `m_Mobile.Spell == null` as equiv opposites, this can be reduced down to:

```suggestion
if (m_Mobile.InRange(c, 1) || m_Mobile.Spell?.IsCasting == true || Core.TickCount - m_Mobile.LastMoveTime > 800)
```

Can we make that assumption?

PR #2197, opened 2025-05-24
fix: Fixes pet validation for resurrection spell

Bohicatv:


Projects/UOContent/Spells/Eighth/Resurrection.cs
```diff
@@ -26,26 +27,30 @@ public ResurrectionSpell(Mobile caster, Item scroll = null) : base(caster, scrol
...
+            if (!Caster.Alive)
+            {
+                Caster.SendLocalizedMessage(501040); // The resurrecter must be alive.
+            }
+            else if (m == Caster)
             {
                 Caster.SendLocalizedMessage(501039); // Thou can not resurrect thyself.
             }
-            else if (!Caster.Alive)
+            else if (m is BaseCreature)
```
kamronbatman, line 38:
```suggestion
            else if (m is BaseCreature { IsDeadBondedPet: true })
```

Projects/UOContent/Spells/Eighth/Resurrection.cs
```diff
@@ -26,26 +27,30 @@ public ResurrectionSpell(Mobile caster, Item scroll = null) : base(caster, scrol
 
         public void Target(Mobile m)
         {
-            if (m == Caster)
+            if (!Caster.Alive)
+            {
+                Caster.SendLocalizedMessage(501040); // The resurrecter must be alive.
+            }
```
kamronbatman: I don't think this ever triggers cause you can't cast any spells while dead?

Projects/UOContent/Spells/Eighth/Resurrection.cs
```diff
@@ -26,26 +27,30 @@ public ResurrectionSpell(Mobile caster, Item scroll = null) : base(caster, scrol
...
+            else if (m == Caster)
             {
                 Caster.SendLocalizedMessage(501039); // Thou can not resurrect thyself.
             }
```
kamronbatman: Since you can't cast spells while alive, you can't resurrect yourself anyways, so this should never trigger either? 🤔

Projects/UOContent/Spells/Eighth/Resurrection.cs
```diff
@@ -26,26 +27,30 @@ public ResurrectionSpell(Mobile caster, Item scroll = null) : base(caster, scrol
...
-            else if (!Caster.Alive)
+            else if (m is BaseCreature { IsDeadBondedPet: true })
             {
-                Caster.SendLocalizedMessage(501040); // The resurrecter must be alive.
+                Caster.SendMessage("Target can not be revived this way.");
```
kamronbatman, line 36: I wonder what the message on OSI is right now?

Bohicatv:
That's actually my fault. I should have been more clear in the initial method used to discover the bug.

Projects/UOContent/Spells/Eighth/Resurrection.cs
```diff
@@ -26,26 +27,30 @@ public ResurrectionSpell(Mobile caster, Item scroll = null) : base(caster, scrol
...
+            }
+            else if (!m.Player)
+            {
+                Caster.SendLocalizedMessage(501043); // Target is not a being.
             }
```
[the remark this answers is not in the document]
kamronbatman: The `!m.Player` is definitely caught by BaseCreature because it is only set to true for newly created aPlayerMobiles.

PR #2202, opened 2025-05-27
fix: Prevents recursion in BaseCreature.DoHarmful()

Bohicatv:


kamronbatman:
I'll be honest, I don't know if this has side effects. I gotta think about it.

kamronbatman:
I changed this to another potential fix to work around the side effect of not force changing the combatant of pet owners.

PR #2219, opened 2025-06-24
fix: Fixes empty/missing bank box prevent withdraw from account gold

kamronbatman:


Projects/UOContent/Mobiles/Townfolk/Banker.cs
```diff
@@ -380,11 +380,9 @@ public override void OnSpeech(SpeechEventArgs e)
                                 {
                                     Say(1048147); // Your backpack can't hold anything else.
                                 }
-                                else if (amount > 0)
+                                else if (amount <= 0)
                                 {
-                                    var box = e.Mobile.FindBankNoCreate();
-
-                                    if (box == null || !Withdraw(e.Mobile, amount))
```
kamronbatman, line 387: Crux of the fix, we don't need a bank check here. It hinders account wide gold checks.

PR #2220, opened 2025-06-26
fix: Fixes control target not reset when giving pet order from context menu

kamronbatman:


Projects/UOContent/Mobiles/AI/BaseAI.cs
```diff
@@ -2975,6 +2971,7 @@ public override void OnClick(Mobile from, IEntity target)
                     {
                         if (bc.CheckControlChance(from))
                         {
+                            bc.ControlTarget = null;
```
kamronbatman, line 2974: Crux of the change.

PR #2254, opened 2025-10-21
ref: Improve performance metrics and OS platform detection

Bohicatv:
Refactor Main.cs to improve performance metrics and OS platform detection.

### DIFF
--- a/Projects/Server/Main.cs
+++ b/Projects/Server/Main.cs
@@ -120,13 +120,7 @@ public static class Core
     {
         get
         {
-            var span = _cyclesPerSecond.AsSpan();
-            if (span.IsEmpty)
-            {
-                return 0.0;
-            }
-            
-            return _cpsSum / span.Length;
+            return _cyclesPerSecond.Length == 0 ? 0 : _cpsSum / _cyclesPerSecond.Length;
         }
     }
 


Projects/Server/Main.cs
```diff
@@ -53,12 +55,21 @@ public static class Core
...
     public static readonly bool IsFreeBSD = RuntimeInformation.IsOSPlatform(OSPlatform.FreeBSD);
     public static readonly bool IsLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || IsFreeBSD;
     public static readonly bool IsBSD = IsDarwin || IsFreeBSD;
-    public static readonly bool Unix = IsBSD || IsLinux;
+    public static readonly bool Unix = UnixPlatforms.Any(RuntimeInformation.IsOSPlatform);
+
+    private static readonly SearchValues<string> XUnitAssemblyNames = 
+        SearchValues.Create(["xunit.core", "xunit.runner", "xunit.extensibility", "xunit"], StringComparison.OrdinalIgnoreCase);
+
+    private static readonly ArrayPool<double> CycleArrayPool = ArrayPool<double>.Create(128, 10);
```
kamronbatman: We can use STArrayPool.Shared here to avoid creating an extra array pool for no reason. This is especially important since MUO is not multithreaded here.

Projects/Server/Main.cs
```diff
@@ -144,86 +198,83 @@ public static string BaseDirectory
...
     public static IEnumerable<string> FindDataFileByPattern(string pattern)
     {
-        var options = new EnumerationOptions { MatchCasing = MatchCasing.CaseInsensitive };
-        foreach (var p in ServerConfiguration.DataDirectories)
+        var options = new EnumerationOptions 
+        { 
+            MatchCasing = MatchCasing.CaseInsensitive,
+            RecurseSubdirectories = false,
+            ReturnSpecialDirectories = false
+        };
```
kamronbatman: We can actually make this static.

Projects/Server/Main.cs
```diff
@@ -345,6 +396,12 @@ private static void HandleClosed()
         {
             EventSink.InvokeShutdown();
         }
+
+        if (_cyclesPerSecond != null)
+        {
+            CycleArrayPool.Return(_cyclesPerSecond);
```
kamronbatman: Are we trying to treat this like a circular buffer? If so, we should just treat it like that. Hold a static array (no need for an array pool), with a start and count pointer.

Projects/Server/Main.cs
```diff
@@ -90,33 +102,75 @@ public static bool IsRunningFromXUnit
...
+        public readonly long Uptime;
+        
+        public PerformanceMetrics(double currentCPS, ReadOnlySpan<double> samples, long uptime)
+        {
+            CurrentCPS = currentCPS;
+            AverageCPS = CalculateAverage(samples);
+            Uptime = uptime;
+        }
+        
+        private static double CalculateAverage(ReadOnlySpan<double> samples)
```
kamronbatman: If we want to avoid having to do a foreach on calculation (which is a good idea), we should add/subtract the incoming and outgoing sample value from a recorded sum, then calculate the average simply requires doing `sum / count`.

Projects/Server/Main.cs
```diff
@@ -144,86 +198,83 @@ public static string BaseDirectory
...
-
     public static bool EJ => Expansion >= Expansion.EJ;
 
     public static string FindDataFile(string path, bool throwNotFound = true)
     {
-        string fullPath = null;
-
-        foreach (var p in ServerConfiguration.DataDirectories)
+        ReadOnlySpan<char> pathSpan = path.AsSpan();
+
```
kamronbatman:
```suggestion

```

Projects/Server/Main.cs
```diff
@@ -144,86 +198,83 @@ public static string BaseDirectory
...
         }
 
-        if (fullPath == null && throwNotFound)
+        if (throwNotFound)
         {
             throw new FileNotFoundException($"Data: {path} was not found");
         }
-
-        return fullPath;
+
```
kamronbatman:
```suggestion

```

Projects/Server/Main.cs
```diff
@@ -90,33 +102,75 @@ public static bool IsRunningFromXUnit
...
+        public PerformanceMetrics(double currentCPS, ReadOnlySpan<double> samples, long uptime)
+        {
+            CurrentCPS = currentCPS;
+            AverageCPS = CalculateAverage(samples);
+            Uptime = uptime;
+        }
+        
+        private static double CalculateAverage(ReadOnlySpan<double> samples)
+        {
+            if (samples.IsEmpty)
```
kamronbatman: This will never be empty since the span is always `_cyclesPerSecond.AsSpan()` which is the length of the array. We should slice the original array if we want it to only be the size of the against the next index.

Projects/Server/Main.cs
```diff
@@ -73,9 +84,10 @@ public static bool IsRunningFromXUnit
                 return _isRunningFromXUnit.Value;
             }
 
-            foreach (var a in AppDomain.CurrentDomain.GetAssemblies())
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
```
kamronbatman, line 87: This work is done once and cached, so we don't have to bother optimizing it.

Projects/Server/Main.cs
```diff
@@ -90,33 +102,53 @@ public static bool IsRunningFromXUnit
...
+    {
+        get
+        {
+            var span = _cyclesPerSecond.AsSpan();
+            if (span.IsEmpty)
+            {
+                return 0.0;
+            }
+            
+            return _cpsSum / span.Length;
```
kamronbatman:
```suggestion
            return _cyclesPerSecond.Length == 0 ? 0 : _cpsSum / _cyclesPerSecond.Length;
```

Projects/Server/Main.cs
```diff
@@ -53,12 +55,21 @@ public static class Core
 
     private static readonly Type[] _serialTypeArray = { typeof(Serial) };
 
+    private static readonly FrozenSet<OSPlatform> UnixPlatforms =
```
kamronbatman, line 58: This is not actually faster than the static computed boolean value. 🤔

Bohicatv: Not really convinced either.

PR #2267, opened 2025-11-15
fix: fixes pets from attacking themselves (Again)

Bohicatv:
Reverts modernuo/ModernUO#2265


kamronbatman:
Let's make the mage only adjustment here.

Bohicatv:
Original issue was a pain to track down with the pet attacking itself, but the main issue was, the attack never went off, so it was hard to figure out what exactly was happening unless I enabled a spell casting target hue change in client to show when the casting of spells were going off even with debugging enabled. So a revert was needed here to clean this up and actually fix the root problem. The revert is mainly to remove the unnecessary additions to the other AI types and to just address the spell portion by blocking spells entire from going off if the pet is targeting itself.

kamronbatman (approved):

PR #2271, opened 2025-11-16
fix: Fixes movement pathing while in combat

Bohicatv:
### Summary

* Removes the `run` flag because it causes most NPC animations to look weird because the mob doesn't have a dedicated run animation.

### DIFF
--- a/Projects/UOContent/Mobiles/AI/ArcherAI.cs
+++ b/Projects/UOContent/Mobiles/AI/ArcherAI.cs
@@ -41,7 +41,7 @@ public class ArcherAI : BaseAI
             return true;
         }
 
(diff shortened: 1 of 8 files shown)


Projects/UOContent/Mobiles/AI/ArcherAI.cs
```diff
@@ -41,13 +41,7 @@ public override bool DoActionCombat()
             return true;
         }
 
-        if (Core.TickCount - Mobile.LastMoveTime > 1000 && !WalkMobileRange(
```
kamronbatman, line 44: Just double checking here, this check that gates the movement by 1 second was removed, was that on purpose?

Bohicatv: Yes. Its not needed.

kamronbatman (approved):

PR #2272, opened 2025-11-16
fix: fixes descending comparer in admin gump

Bohicatv:


Projects/UOContent/Gumps/AdminGump.cs
```diff
@@ -4087,23 +4087,23 @@ obj is NetState ns &&
             }
         }
 
-        private class SharedAccountDescendingComparer : IComparer<object>
+        public class SharedAccountDescendingComparer : IComparer<object>
```
kamronbatman: Do we need this to be public?

PR #2275, opened 2025-11-18
perf: uses span based GetMobilesInRange

Bohicatv:


Projects/Server/Mobiles/Mobile.cs
```diff
@@ -8061,6 +8061,46 @@ public Map.MobileAtEnumerable<T> GetMobilesInRange<T>() where T : Mobile =>
     public Map.MobileBoundsEnumerable<T> GetMobilesInRange<T>(int range) where T : Mobile =>
         m_Map == null ? Map.MobileBoundsEnumerable<T>.Empty : m_Map.GetMobilesInRange<T>(m_Location, range);
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void GetMobilesInRange<T>(int range, Span<T> buffer, out int count) where T : Mobile
```
kamronbatman, line 8065: We should return `int` instead of out, similar to `int Read(Span<byte>)`

Projects/Server/Mobiles/Mobile.cs
```diff
@@ -8061,6 +8061,46 @@ public Map.MobileAtEnumerable<T> GetMobilesInRange<T>() where T : Mobile =>
...
+        }
+
+        var tempList = new List<(T mobile, double distance)>();
+        foreach (var mobile in m_Map.GetMobilesInRange<T>(m_Location, range))
+        {
+            var distance = GetDistanceToSqrt(mobile);
+            tempList.Add((mobile, distance));
+        }
+
+        tempList.Sort((a, b) => a.distance.CompareTo(b.distance));
```
kamronbatman: We shouldn't bother sorting. Instead we can use a maximum range, reference that entity index, and swap it if we get one that is closer, but only if we are close enough. I think we should leave sorting up to the implementer especially since the function doesn't say it sorts by range?

Projects/Server/Mobiles/Mobile.cs
Bohicatv: I think `out` is fine here. More obvious on fill and count. I want to also avoid an extra allocation for the return value. if the method was about "how many" I'd do int.

Bohicatv:
Too many merge conflicts.

PR #2287, opened 2025-12-04
fix: fixes pet guard order

Bohicatv:


Projects/UOContent/Mobiles/AI/BaseAI/PetOrders.cs
```diff
@@ -340,23 +342,57 @@ private void HandleInvalidControlTarget()
...
+                if (Mobile.InLOS(aggr))
+                {
+                    Mobile.ControlTarget = aggr;
+                    Mobile.ControlOrder = OrderType.Attack;
+                    Mobile.Combatant = aggr;
+
+                    var target = isAttackingMaster ? "master" : "me";
+                    this.DebugSayFormatted($"{aggr.Name} is attacking my {target}! Engaging...");
+
+                    Think();
```
kamronbatman, line 369: We have to be careful that this doesn't result in the mob being damaged, deleted, moved, etc. Should we be defensive and use a PooledRefQueue?

Projects/UOContent/Mobiles/AI/BaseAI/PetOrders.cs
```diff
@@ -340,23 +342,57 @@ private void HandleInvalidControlTarget()
...
+                
+                if (Mobile.InRange(aggressor, Mobile.RangePerception) && Mobile.CanSee(aggressor) && Mobile.InLOS(aggressor))
+                {
+                    Mobile.ControlTarget = aggressor;
+                    Mobile.ControlOrder = OrderType.Attack;
+                    Mobile.Combatant = aggressor;
+
+                    this.DebugSayFormatted($"{aggressor.Name} recently attacked my master! Retaliating...");
+
+                    Think();
```
kamronbatman, line 394: Same, is it possible that Aggressors gets modified while we are iterating? Here we might be able to get away with iterating backwards?

Projects/UOContent/Mobiles/AI/BaseAI/PetOrders.cs
Bohicatv: Aah! for sure, forgot about the potential crash here. But I don't want to do pooledref, because that would be a lot of overhead if a lot of pets are on guard. How likely? I dunno. Maybe if I condition the mobile counts to like 20.

Projects/UOContent/Mobiles/AI/BaseAI/PetOrders.cs
```diff
@@ -340,23 +345,73 @@ private void HandleInvalidControlTarget()
...
+                {
+                    Mobile.ControlTarget = aggr;
+                    Mobile.ControlOrder = OrderType.Attack;
+                    Mobile.Combatant = aggr;
+
+                    var target = isAttackingMaster ? "master" : "me";
+                    this.DebugSayFormatted($"{aggr.Name} is attacking my {target}! Engaging...");
+
+                    Think();
+                    return;
```
kamronbatman, line 370: I didn't see this return at first, I think we can avoid the pooled ref because this is terminal.

kamronbatman (approved):

PR #2348, opened 2026-03-02
fix: Fixes criminals having guards called on them

kamronbatman:
## Summary

Fixes three bugs in `GuardedRegion.CallGuards`:

- **Operator precedence bug**: The condition `!m.Region.IsPartOf(this) && !m_GuardCandidates.ContainsKey(m)`
  was inverted from intent. Replaced with explicit split: dictionary members are targeted regardless of
  region; permanent candidates (reds/AlwaysMurderer) must be inside the region.
- **Premature `break`**: Only the first guard candidate was ever processed per "guards" call.
  Removed the `break` so all valid candidates in range get a guard spawned.
- **Misleading message for permanent reds**: "Guards can no longer be called on you." (502276)
  was sent to permanent reds, but guards can *always* be called on them. Now only sent to
  temporary criminals whose guard window is actually consumed.

Also extracts `IsAlwaysGuardCandidate()` helper and simplifies `IsGuardCandidate()`.

## Edge cases verified (by analysis)

- **Multiple players call guards on same red**: `BaseGuard.Spawn` dedup (scans 15 tiles for
  existing guard with same `Focus`) prevents duplicate guards. Message suppression eliminates spam.
- **Red fights spawned guard → criminal → guards called again**: Same dedup prevents infinite
  guard spawns. Existing guard already has `Focus == red`.

## Test plan

- [ ] Temporary criminal in guarded region → guard spawns, receives "Guards can no longer be called on you."
- [ ] Permanent red (5+ kills) in guarded region → guard spawns, does NOT receive the message
- [ ] Multiple players call "guards" on same red → only 1 guard spawns
- [ ] Criminal outside region but in dictionary → still targeted by guards call from inside region
- [ ] Red outside guarded region → NOT targeted by guards call (must be inside region)

### DIFF
--- a/Projects/UOContent/Regions/GuardedRegion.cs
+++ b/Projects/UOContent/Regions/GuardedRegion.cs
@@ -135,7 +135,7 @@ public class GuardedRegion : BaseRegion
     }
 
     public virtual bool CheckVendorAccess(BaseVendor vendor, Mobile from) =>
-        from.AccessLevel >= AccessLevel.GameMaster || IsDisabled() || from.Kills < 5;
+        from.AccessLevel >= AccessLevel.GameMaster || IsDisabled() || !from.Murderer;
 
     public override bool OnBeginSpellCast(Mobile m, ISpell s)
     {
@@ -171,7 +171,7 @@ public class GuardedRegion : BaseRegion
             return;
         }
 
-        if (!AllowReds && (m.Kills >= 5 || m is BaseCreature { AlwaysMurderer: true }))
+        if (!AllowReds && m.Murderer)
         {
             CheckGuardCandidate(m);
         }
@@ -332,8 +332,7 @@ public class GuardedRegion : BaseRegion
         }
     }
 
-    private bool IsAlwaysGuardCandidate(Mobile m) =>
-        !AllowReds && (m.Kills >= 5 || (m as BaseCreature)?.AlwaysMurderer == true);
+    private bool IsAlwaysGuardCandidate(Mobile m) => !AllowReds && m.Murderer;
 
     public bool IsGuardCandidate(Mobile m)
     {
@@ -342,7 +341,7 @@ public class GuardedRegion : BaseRegion
             return false;
         }
 
-        return (m as BaseCreature)?.IsInvulnerable != true && (!AllowReds && IsAlwaysGuardCandidate(m) || m.Criminal);
+        return (m as BaseCreature)?.IsInvulnerable != true && (IsAlwaysGuardCandidate(m) || m.Criminal);
     }
 
     private class GuardTimer : Timer
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,7 @@
     <PropertyGroup>
         <Authors>Kamron Batman</Authors>
         <Company>ModernUO</Company>
-        <Copyright>2019-2025</Copyright>
+        <Copyright>2019-2026</Copyright>
         <TargetFramework>net10.0</TargetFramework>
         <LangVersion>14</LangVersion>
         <PublicRelease>true</PublicRelease>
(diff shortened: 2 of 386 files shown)


Projects/UOContent/Regions/GuardedRegion.cs
```diff
@@ -298,43 +302,47 @@ public void CallGuards(Point3D p)
...
+
+            if (!IsAlwaysGuardCandidate(m))
+            {
+                m.SendLocalizedMessage(502276); // Guards can no longer be called on you.
+            }
         }
     }
 
+    private bool IsAlwaysGuardCandidate(Mobile m) =>
+        !AllowReds && (m.Kills >= 5 || (m as BaseCreature)?.AlwaysMurderer == true);
```
jackuoll: Note: this is bit of a ramble not directly related to this PR. I am happy to open an issue and work on everything mentioned here.

-----------

Just a question/observation - AlwaysMurderer is in BaseCreature but during T2A/UOR, players could basically have the "AlwaysMurderer" flag also in the form of ping pongs. If I can make a suggestion, we could simplify this to avoid people tripping up on "murderer gotchas" by implementing a single definition. Below is a ramble about a potential implementation that I'd be happy to do.

## Ping Pongs and simplficiation of IsMurderer

### What are ping pongs? 

A player who has gone blue -> red had their ping pong counter incremented by 1. On 5 ping pongs, the player would be permanently red, even if their kill counter dropped below 5.

### ModernUO potential implementation/simplficiation of IsMurderer

There are checks all over the codebase to determine Kills >= 5. This has always been the case, dating back to RunUO. Everyone is doing their own individual checks to figure out what a murderer is. I would be happy to implement pingpongs and add a Murderer property into Mobile.cs to simplify this whole system and stop the gotchas with murderer flags.

IMO this would include:
* Add a ping/pong counter to PlayerMobile.cs
* Add a virtual IsMurderer property into Mobile.cs that defaults to Kills >= 5
  * Override in BC with base.IsMurderer || AlwaysMurderer
  * Override in PM with base.IsMurderer || PingPong >= 5
* Replace Kills >= 5 and AlwaysMurderer checks over the entire codebase with IsMurderer.
* Add ping/pong into the "I must consider my sins" speech command.

### Era accuracy

The last mention of ping pongs in a patch note I could find was publish 12, which was not a removal patch note, it was just informational. I checked all the way through to 2005 and did not find another mention of ping pongs after that.

kamronbatman: Hey @jackuoll - I agree with you that `.Kills > 5` and the concept of `IsMurderer` (red) needs to be consolidated. Regarding `AlwaysMurderer` for players, I believe this was the mechanic _before shorts_ were added, and I believe it was even before T2A.

Regarding your implementation proposal, I agree and we can definitely do that. For the ping-pong counter itself, I am not too sure how that worked, but if ideas please let me know.

We can the ping-pong counter to the PlayerMurderSystem (ModernUO decoupled murder counts themselves outside of Kills) and put that behind a flag in modernuo.json and/or the FeatureFlag system.

Regarding "I must consider my sins", we can definitely implement whatever is era accurate. Here are the clilocs
`502122` -> Fear not, thou hast not slain the innocent.
`502123` -> Thou art known throughout the land as a murderous brigand.
`502124` -> Fear not, thou hast not slain the innocent in some time...
`502125` -> Although thou hast slain the innocent, thy deeds shall not bring retribution upon thy return to the living.
`502126` -> If thou should return to the land of the living, the innocent shall wreak havoc upon thy soul.


I recommend we open an issue about this to track the work to implement these changes.

jackuoll: Added https://github.com/modernuo/ModernUO/issues/2352 for the pingpong/IsMurderer change.

I actually edited out the part about "I must consider my sins" as this is T2A but not UOR. I'm not even sure if there is any real T2A support across the codebase currently, but I'd be happy to create an issue and work on it. I'll flag it for now.

kamronbatman: I do want to add T2A/UOTD support.

jackuoll: T2A accuracy can be pretty opinionated! I think UOSA is the gold standard, but there's a lot of things that were never implemented there. The cutoff for UOSA is Nov 23, 1999, but they pick and choose features from that date. How accurate do you want T2A? I'd be happy to chat about this as I have spent a _lot_ of time in my life researching T2A (and implementing T2A features)

PR #2453, opened 2026-05-13
Add Spectral Spellbinder for Old Haven

creiht:
* Validated that it spawns in Old Haven
* Validated that it only attacks if attacked
* Validated that it doesn't cast damage based spells
* Used ServUO and [UOGuide](https://www.uoguide.com/Spectral_Spellbinder) for reference

### DIFF
--- a/Projects/UOContent/Migrations/Server.Mobiles.Spellbinder.v0.json
+++ b/Projects/UOContent/Migrations/Server.Mobiles.Spellbinder.v0.json
@@ -1,0 +1,4 @@
+{
+  "version": 0,
+  "type": "Server.Mobiles.Spellbinder"
+}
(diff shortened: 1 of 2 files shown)


kamronbatman:
Thanks!

This mob only casts non-damage attack spells and melee otherwise. It will require a specific Mage AI variant or otherwise restrict the spells it can cast.

creiht:
> Thanks!
> 
> This mob only casts non-damage attack spells and melee otherwise. It will require a specific Mage AI variant or otherwise restrict the spells it can cast.

What would you think about using a Mage AI subclass and overriding GetRandomDamageSpell() to return GetRandomCurseSpell() (or it could also just return `null` to do nothing?  Or is that too janky?

kamronbatman:
> > Thanks!
> > This mob only casts non-damage attack spells and melee otherwise. It will require a specific Mage AI variant or otherwise restrict the spells it can cast.
> 
> What would you think about using a Mage AI subclass and overriding GetRandomDamageSpell() to return GetRandomCurseSpell() (or it could also just return `null` to do nothing? Or is that too janky?

We could do thet. I wonder if we need a new AI type enum too. Feel free to look into options.

creiht:
I updated it so that the creature will cast a curse spell rather than a damage based spell.

PR #2459, opened 2026-05-30
fix(ai): pet order/home refactor — stop & post-combat behavior

kamronbatman:
## Summary

Fixes two related pet-behavior bugs and the underlying design flaw behind both:

1. **Post-combat erratic** — after a pet killed its `all kill` target it milled around erratically at the kill site (or failed to return to the master) until the player issued `all follow`/`all stop`.
2. **`all stop` returns home** — a pet with a non-zero `Home` walked back toward that location on `all stop` (on ML; on non-ML the old `DoOrderStop` was a no-op, so the sighting there came from a residual `Stay`).

### Root cause

`ControlOrder` and the wild-creature `Home` field were overloaded to express several distinct ideas, mutated/read inconsistently across order transitions:

- `Home` doubled as the controlled-pet "stay anchor" (`HandleStayOrder` set `Home = Location`) but nothing cleared it when the pet left the staying state; `HandleStopOrder` was the only handler that never touched it.
- `DoOrderStop` had dropped RunUO's `Home = Location` re-anchor, so on ML it walked to a stale anchor.
- The post-combat fallback was a fragile `_lastPetOrder` hack in `DoOrderNone` that re-anchored a resumed `Stay` at the corpse.
- Controlled idle-wander bypassed the `CheckIdle()` rest gate that every non-controlled creature uses, so idling pets jittered every AI tick.

## Approach

Separate three concepts that were tangled together:

- **`ControlOrder`** — the active order (may be transient: Come/Attack/Drop).
- **Persistent command** (`PersistentOrder` ∈ `{None, Stay, Follow, Guard}`) — the standing directive a pet falls back to when a transient order completes. Runtime-only (not serialized; reset to `None` on load) and **derived from master proximity on login** (near → Follow, far → Stay).
- **Anchor** (`Home`) — a pure function of the persistent command, set only when that command changes (never on transient transitions or fallback-resume), so it can't go stale.

### Behavior

- **Stop** is resolved immediately from what the pet was doing: Attack/Come → resume the persistent command; Follow/Guard → cancel to idle where it stands; Stay → stay put.
- **Stay** holds its post (returns only if displaced, e.g. after a fight) — no shuffle.
- **Idle** (`None`) is a gentle wander routed through `CheckMove/CanMoveNow/CheckIdle`, so idling pets take the same 15–25s rest periods as other creatures, on both ML and non-ML.
- **Post-combat** the pet resumes its persistent command (a staying pet returns to its original post, not the corpse).
- **Release** without a spawner anchors where the pet stands instead of pathing to a stale anchor.

This restores the RunUO-intended behavior (verified against the RunUO reference) while fixing the ModernUO regressions.

## Tests

New `PetOrderTests` (13 deterministic xUnit tests) cover: anchor lifecycle, the full Stop truth table, report 1 (post-combat return to post), report 2 (no stale-anchor walk-home), frozen-Stay/gated-idle wiring, release fix, derive-on-login, and a non-ML spot-check. The subjective wander *feel* is covered by a manual-QA checklist in the implementation plan.

## Notes

- Engine project (`Projects/Server`) untouched; the one `BaseCreature.cs` change is the `ControlOrder` setter passing the previous order to `OnCurrentOrderChanged`.
- `DoOrderCome` keeps auto-converting to `Stay` on arrival, which under the new model cleanly means "come and hold near me."
- Commits in this PR are temporarily **unsigned** (the signing agent's passphrase cache expired mid-session); happy to re-sign / amend on request.

### DIFF
--- a/Projects/UOContent/Mobiles/AI/BaseAI/PetLogin.cs
+++ b/Projects/UOContent/Mobiles/AI/BaseAI/PetLogin.cs
@@ -1,0 +1,50 @@
+/*************************************************************************
+ * ModernUO                                                              *
+ * Copyright 2019-2026 - ModernUO Development Team                       *
+ * Email: hi@modernuo.com                                                *
+ * File: PetLogin.cs                                                     *
+ *                                                                       *
+ * This program is free software: you can redistribute it and/or modify  *
+ * it under the terms of the GNU General Public License as published by  *
+ * the Free Software Foundation, either version 3 of the License, or     *
+ * (at your option) any later version.                                   *
+ *                                                                       *
+ * You should have received a copy of the GNU General Public License     *
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.  *
+ ************************************************************************/
+
+using ModernUO.CodeGeneratedEvents;
+
+namespace Server.Mobiles;
+
+public static class PetLoginHandler
+{
+    // Within this many tiles of the master we assume the pet was following; otherwise staying.
+    private const int FollowRange = 12;
+
+    [OnEvent(nameof(PlayerMobile.PlayerLoginEvent))]
+    public static void OnLogin(PlayerMobile pm) => DeriveFollowerOrders(pm);
+
+    // The persistent command is runtime-only and reset to None on load. When the master logs
+    // in we give each controlled pet that still has no standing command a sane one, inferred
+    // from proximity: near master -> Follow, otherwise Stay.
+    public static void DeriveFollowerOrders(PlayerMobile master)
+    {
+        if (master?.AllFollowers == null)
+        {
+            return;
+        }
+
+        foreach (var follower in master.AllFollowers)
+        {
+            if (follower is BaseCreature { Controlled: true, Deleted: false } bc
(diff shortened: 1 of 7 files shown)


Projects/UOContent/Mobiles/AI/BaseAI/PetLogin.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Server.Mobiles;
+
+public static class PetLoginHandler
+{
+    // Within this many tiles of the master we assume the pet was following; otherwise staying.
+    private const int FollowRange = 12;
+
+    [OnEvent(nameof(PlayerMobile.PlayerLoginEvent))]
+    public static void OnLogin(PlayerMobile pm) => DeriveFollowerOrders(pm);
```
kamronbatman, line 26: This is doing extra work if we just auto claimed pets. See https://github.com/modernuo/ModernUO/blob/b589da3efb20b7b89395995ba2762b919d78ca03/Projects/UOContent/Mobiles/PlayerMobile.cs#L1283

PR #2473, opened 2026-06-07
fix(tests): single shared bootstrap; idempotent SerializationThreadWorker.Exit

kamronbatman:
## Problem

Running the full `UOContent.Tests` suite, the test host **hangs ~2.5 minutes at shutdown and then crashes** (`Test host process crashed` / run aborted). The tests themselves are fine — they complete in ~1s — but the process can't exit.

Captured via `--blame-hang` dump. The blocking thread:

```
System.Threading.WaitHandle.WaitOne()
Server.SerializationThreadWorker.Sleep()        SerializationThreadWorker.cs:54  (_stopEvent.WaitOne())
Server.SerializationThreadWorker.Exit()         SerializationThreadWorker.cs:61
Server.World.ExitSerializationThreads()         World.cs:429
Server.Tests.UOContentFixture..ctor()
```

### Root cause

Both collection fixtures (`UOContentFixture` and `PathfindingTestFixture`) each run the **full process-global ModernUO bootstrap**. `World.Load()` is guarded to run once per process, so the **second** fixture's `World.Load()` is a no-op and does **not** respawn the serialization workers — but `World.ExitSerializationThreads()` is **not** guarded, so the second fixture calls `Exit()` on workers whose threads have already terminated. `Exit()` → `Sleep()` → `_stopEvent.WaitOne()` then blocks forever (a dead thread never sets the event). The first collection's tests run; the second collection's fixture deadlocks in its constructor; the host eventually gets killed.

This is why single-collection (filtered) runs were fine — only one fixture ever bootstraps — but the full suite hangs. It's not a parallelization race: even strictly sequential, the second fixture deadlocks.

## Fix

**(a) Engine — idempotent `SerializationThreadWorker.Exit()`**
A second `Exit()` is now a safe no-op instead of a permanent block. Only the owning (main) thread calls `Exit()`, so no synchronization is needed, and the single-call production shutdown path is unchanged.

**(b) Tests — one shared bootstrap, strictly sequential collections**
- New `TestServerBootstrap.EnsureInitialized()` runs the superset global init **exactly once per process** (lock + once-flag).
- `UOContentFixture` / `PathfindingTestFixture` slim down to delegate to it and no longer tear down global state (which the single-bootstrap model owns for the host's lifetime).
- `[assembly: CollectionBehavior(DisableTestParallelization = true)]` so collections never overlap.

## Result

| | Before | After |
|---|---|---|
| Tests run (full suite) | 258 (UOContent collection deadlocked) | **418** |
| Outcome | 2.5-min hang → host crash | **418 passed, clean exit** |
| Wall time | killed | **~7s** |

### DIFF
--- a/Projects/Server/Serialization/SerializationThreadWorker.cs
+++ b/Projects/Server/Serialization/SerializationThreadWorker.cs
@@ -29,6 +29,7 @@ public class SerializationThreadWorker
     private readonly AutoResetEvent _stopEvent; // Main thread waits for the worker finish draining
     private bool _pause;
     private bool _exit;
+    private bool _exited;
     private byte[] _heap;
 
     private readonly ConcurrentQueue<IGenericSerializable> _entities;
@@ -56,6 +57,15 @@ public class SerializationThreadWorker
 
     public void Exit()
     {
+        // Idempotent: once the worker thread has exited, its _stopEvent is never set again,
+        // so a second Exit() would block forever in Sleep(). Only the worker's owning (main)
+        // thread calls Exit(), so no synchronization is needed on _exited.
+        if (_exited)
+        {
+            return;
+        }
+
+        _exited = true;
         _exit = true;
         Wake();
         Sleep();
--- a/Projects/Server/Serialization/SerializationThreadWorker.cs
+++ b/Projects/Server/Serialization/SerializationThreadWorker.cs
@@ -57,9 +57,6 @@ public class SerializationThreadWorker
 
     public void Exit()
     {
-        // Idempotent: once the worker thread has exited, its _stopEvent is never set again,
-        // so a second Exit() would block forever in Sleep(). Only the worker's owning (main)
-        // thread calls Exit(), so no synchronization is needed on _exited.
         if (_exited)
         {
             return;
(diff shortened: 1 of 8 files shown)


Projects/Server/Serialization/SerializationThreadWorker.cs
```diff
@@ -56,6 +57,15 @@ public void Sleep()
 
     public void Exit()
     {
+        // Idempotent: once the worker thread has exited, its _stopEvent is never set again,
+        // so a second Exit() would block forever in Sleep(). Only the worker's owning (main)
+        // thread calls Exit(), so no synchronization is needed on _exited.
```
kamronbatman:
```suggestion
```

Projects/UOContent.Tests/Fixtures/TestServerBootstrap.cs
```diff
@@ -0,0 +1,121 @@
...
+/// Each xUnit collection has its own fixture instance, so without this shared guard every
+/// collection would re-run the full bootstrap. This runs the superset bootstrap (the union of
+/// what every collection needs, including the UO client data directory and TileData load)
+/// exactly once, no matter how many collection fixtures are constructed. Combined with
+/// <c>[assembly: CollectionBehavior(DisableTestParallelization = true)]</c>, collections run
+/// strictly sequentially and share one initialized world.
+/// </summary>
+internal static class TestServerBootstrap
+{
+    private static readonly object _sync = new();
```
kamronbatman:
```suggestion
    private static readonly Lock _sync = new();
```

PR #2479, opened 2026-06-09
feat(pathfinding): multi-aware mask synthesizer + warm interior cache for house/boat cells

kamronbatman:
## Problem

Houses and boats (multis) were pathed correctly only by **delegation to the slow path**: `StepCache.TryGetMask` returns `Fallthrough_Multi` for any multi-covered cell, and `GetSuccessors` ran `CheckMovement` **8× per cell** (each re-resolving the tile stack via `GetStaticAndMultiTiles`) — a sustained per-step cost near every house/boat. There was also no automated test pinning multi pathfinding.

This branch is the full multi-pathfinding effort in phases on one branch.

## Phase 1 — characterization tests (the oracle)

Implementation-agnostic invariants: a cache-on≡cache-off whole-path invariant, a per-cell sweep vs `CheckMovement` over footprint+halo (incl. destination Z), hand-verified routing (around walls, demolish-reopens, foundation-redesign-honored), classic-house / foundation / boat fixtures, non-vacuity guards. These gate every later phase byte-for-byte.

## Phase 2 — live single-pass synthesizer

`StepProbe.ComputeMultiMaskAt` synthesizes a covered cell's full 8-direction `StepMask` in one pass (the existing surface/step logic over `GetStaticAndMultiTiles` instead of 8× `CheckMovement`). `GetSuccessors` routes `Fallthrough_Multi` cells through it. No new cache, no `.swb` change. **~1.5×**, zero added allocations.

## Phase 3 / 3.1 — warm per-`multiID` interior cache (airtight)

`MultiMaskCache` caches each fixed multi's local-frame `StepMask` for **interior** cells (cell + all 8 neighbours covered → terrain-neighbour-free → position-invariant), keyed by `multiID & 0x3FFF`, built lazily from the MCL. Interior cells become ~20 ns lookups.

The cache is gated on a **per-instance footprint-clean flag** (`BaseMulti.PathInteriorCacheState`): an instance whose whole footprint terrain is below its floor (`maxTerrain < minFloor`) serves from the cache; a **dirty** instance (terrain intrudes — a contrived/GM placement) **degrades to live-synth, never a wrong mask**. This closes a cross-instance soundness gap (the cached mask depends on neighbour terrain too) found in a holistic review. The gate resets whenever the footprint's world-terrain relationship can change — **location, map, or ItemID** (a boat's heading swaps the MCL).

**Boats are cached too.** Their per-`multiID` deck masks are movement-invariant (built once per heading), so a sailing boat never rebuilds them; only the cheap clean-flag rescan repeats per move (and only when pathed near). Narrow existing boats have little interior; wide galleons (`multi.mul`) would gain Castle-class. `HouseFoundation` (per-instance runtime `DesignState`) is the one type that stays on the live path.

## Verification

- `UOContent.Tests` **454/454**, `Server.Tests` **708/708**, 0 failures.
- The Phase-1 oracle (`MultiPathInvariantTests`, cache-on ≡ cache-off) stays **byte-identical** with the synthesizer + interior cache active.
- Tests pin: footprint-cleanliness (clean vs sunk), dirty/cluttered placement degrades to live-synth while still pathing, clean placement serves, and the gate resets on move/ItemID change.

## Performance (modernuo/ModernUO-Benchmarks#8, full-fixture)

Houses at **Green Acres** (flat staff region → clean footprints, the legit-placement case):

| Route | Slow path | Phase 3.1 (interior cache) | Speedup |
|-------|----------:|---------------------------:|--------:|
| `around_a` (29 steps) | 238.3 µs | **49.1 µs** | **4.85×** |
| `around_b` (29 steps) | 224.3 µs | **49.5 µs** | **4.53×** |

~130 of ~167 multi cells/route serve from the cache (~20 ns) vs 37 live-synth. Per-cell, the slow path's 8× `CheckMovement` grows with multi complexity (GuildHouse ~857 ns → Castle ~1,194 ns), the synthesizer is a flat ~780 ns, and the cache serve is ~20 ns — so big/tall multis (and wide galleons) gain most. Identical allocations throughout.

### DIFF
--- a/Projects/UOContent/Engines/Pathing/Cache/MultiMaskCache.cs
+++ b/Projects/UOContent/Engines/Pathing/Cache/MultiMaskCache.cs
@@ -100,6 +100,32 @@ public sealed class MultiMaskCache
         return true;
     }
 
+    /// <summary>
+    /// True iff all terrain (land + statics) at (x,y) sits strictly below <paramref name="floorZ"/>,
+    /// so a creature standing on the multi floor never sees terrain in its envelope and the cached
+    /// (terrain-free) mask is exact. Cheap: one land-top read + the cell's static-tile array scan.
+    /// </summary>
+    public static bool TerrainTopBelow(Map map, int x, int y, sbyte floorZ)
+    {
+        map.GetAverageZ(x, y, out _, out _, out var landTop);
+        if (landTop >= floorZ)
+        {
+            return false;
+        }
+
+        foreach (var tile in map.Tiles.GetStaticTiles(x, y))
+        {
+            var data = TileData.ItemTable[tile.ID & TileData.MaxItemValue];
+            var top = tile.Z + data.CalcHeight;
+            if (top >= floorZ)
+            {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
     /// <summary>Inverse of <see cref="TryToLocalZ"/>: add multiZ back to recover world Zs.</summary>
     public static StepMask ToWorldZ(StepMask local, int multiZ)
     {
--- a/Projects/UOContent/Engines/Pathing/Cache/MultiMaskCache.cs
+++ b/Projects/UOContent/Engines/Pathing/Cache/MultiMaskCache.cs
@@ -46,6 +46,16 @@ public sealed class MultiMaskCache
             // Reconstruct the floor Z in the world frame for THIS placement (int; avoids a silent
             // sbyte wrap when a different instance serves a cell cached at another instance's Z).
             var worldFloorZ = local.FloorZAt(lx, ly) + multi.Z;
+            // KNOWN LIMITATION (cross-instance neighbour terrain): the cached mask is the multi's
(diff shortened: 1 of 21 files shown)


Projects/UOContent/Engines/Pathing/Cache/MultiMaskCache.cs
```diff
@@ -0,0 +1,346 @@
...
+/// dirty instances, boats (movers), and HouseFoundation (runtime-mutable) fall back to live-synth.
+/// Keyed by multiID &amp; 0x3FFF.
+/// </summary>
+public sealed class MultiMaskCache
+{
+    public static MultiMaskCache Instance { get; } = new();
+
+    private const int StepHeight = 2;
+
+    private readonly Dictionary<int, MultiLocalMask> _byMultiId = new();
```
kamronbatman:
```suggestion
    private readonly Dictionary<int, MultiLocalMask> _byMultiId = [];
```

PR #2496, opened 2026-06-19
refactor(archery): centralize SE ammo auto-recovery off PlayerMobile

kamronbatman:
## Summary

Streamlines the SE-era archery ammo auto-recovery (recovering spent arrows/bolts after a miss). The mechanic was previously spread across four unrelated trigger points and a weapon-scoped timer that was divorced from the recovery state living on `PlayerMobile`. It also contained dead code.

### Problems fixed
- **Dead `!Warmode` gate** — `OnMiss` only runs from `OnSwing`, which requires warmode to fire, so the `if (!pm.Warmode)` branch that started the recovery timer could never trigger.
- **Scattered, divorced state** — banked ammo lived on `PlayerMobile.RecoverableAmmo` while the timer lived on the weapon (`_recoveryTimerToken`), and recovery was kicked off from four different places (`OnWarmodeChanged`, `PlayerMobile.OnDamage` kill, `BaseCreature.OnDamage` kill, `OnBeforeDeath`).
- **Per-player footprint** — every `PlayerMobile` carried a `RecoverableAmmo` field even though ~99% never miss with a bow (and most are offline).

### New design — `AmmoRecovery` side table
- All state (banked ammo + one repeating timer) is keyed by player in a static dictionary, so only players who actually miss carry any state. Transient by design — this was never serialized.
- **One feed point:** `OnMiss` banks the spent ammo type and starts the player's timer.
- **One drain point:** the timer self-gates and gathers ammo into the backpack only once the archer has disengaged — **alive, out of warmode, and not running** — otherwise it retries next tick, so banked ammo is never lost while online.
- **"Not running" allows standing still _or_ walking.** The `Direction.Running` bit is stale after a player stops, so it's paired with movement recency (`LastMoveTime`); only an *actively* running archer is blocked.
- Removed the redundant scattered triggers, the dead `!Warmode` branch, `RecoverableAmmo`, `RecoverAmmo()`, and the now-empty `OnWarmodeChanged` override. `PlayerMobile.OnDelete` calls `AmmoRecovery.Forget`.

### Behavior notes
- On death, banked ammo is **no longer flushed to the corpse** — it stays banked and is recovered after resurrection once the archer settles (player keeps it rather than dropping it to looters).
- `OnHit` immediate recovery (the ~40% arrow-to-defender behavior) is **unchanged**.

## Test plan
- [x] `dotnet build Projects/UOContent/UOContent.csproj -c Release` — succeeds, 0 warnings, 0 errors.
- [ ] In-game (SE era): miss bow shots, then disengage (drop warmode + stop) and confirm the "You recover N arrows/bolts" message and backpack contents; confirm recovery does **not** fire while running and **does** while walking/standing.

### DIFF
--- a/Projects/UOContent/Items/Weapons/Ranged/AmmoRecovery.cs
+++ b/Projects/UOContent/Items/Weapons/Ranged/AmmoRecovery.cs
@@ -1,0 +1,144 @@
+using System;
+using System.Collections.Generic;
+using Server.Mobiles;
+
+namespace Server.Items
+{
+    /// <summary>
+    /// SE-era archery ammo recovery. When a player misses a shot there is a chance the spent ammo
+    /// can be gathered back up afterwards. The banked ammo is held off of <see cref="PlayerMobile"/>
+    /// (in this side table) so the vast majority of players who never miss with a bow carry no extra
+    /// state. A single repeating timer per tracked player gathers the ammo, but only once the archer
+    /// has disengaged: out of warmode and not running (standing still or walking is fine).
+    /// </summary>
+    public static class AmmoRecovery
+    {
+        // How often we re-check whether the archer has settled enough to gather their ammo.
+        private static readonly TimeSpan RecoveryInterval = TimeSpan.FromSeconds(10);
+
+        // A running step is only "current" if one landed this recently. Comfortably above the running
+        // step cadence (run-foot ~200ms, run-mount ~100ms) so an actively running archer always reads as
+        // running, while a stopped runner ages out within ~half a second.
+        private const long RunningWindowMillis = 500;
+
+        // Only players who have banked recoverable ammo appear here, so this stays tiny.
+        private static readonly Dictionary<PlayerMobile, RecoveryState> _states = new();
+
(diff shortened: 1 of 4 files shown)


Projects/UOContent/Items/Weapons/Ranged/AmmoRecovery.cs
```diff
@@ -0,0 +1,144 @@
...
+                }
+
+                ammo.Amount = amount;
+
+                var name = ammo.Name ?? ammo switch
+                {
+                    Arrow _ => $"arrow{(ammo.Amount != 1 ? "s" : "")}",
+                    Bolt _  => $"bolt{(ammo.Amount != 1 ? "s" : "")}",
+                    _       => $"#{ammo.LabelNumber}"
+                };
```
kamronbatman: if LabelNumber is 1020000 + 3903 or 3904 (arrow) but amount is not 1, then increase by 2 to get the plural. For bolts it is 1020000 + 7163 or 7164, same as arrow.

This means we should always have $"#{label}" for all scenarios where `Name` is null.

PR #2506, opened 2026-06-26
fix(json): Rectangle3DConverter loses a z-level on write

kamronbatman:
## Summary

`Rectangle3DConverter.Write` corrupts a rectangle's Z range for certain bounds. The omit-z guard was:

```csharp
var writeZ = value.Start.Z is > sbyte.MinValue and < sbyte.MaxValue
          || value.End.Z   is > sbyte.MinValue and < sbyte.MaxValue;
```

This drops `z1`/`z2` whenever **both** Z bounds sit at/outside the sbyte extremes — but `Read` reconstructs absent Z as exactly `z1 = -128, z2 = 127` (depth 255). So any rectangle that trips the omit condition without *being* that sentinel round-trips to depth 255 and is corrupted.

The concrete case: a **homeRange-style** bound `Start.Z = -128, End.Z = 128` (depth 256, the full vertical range used by spawners) gets `z1`/`z2` omitted on write, then reads back as depth **255** — silently losing the top z-level on every re-serialize.

(Spotted while working on #2505; spawners now prefer the `homeRange` form so most square bounds avoid this path, but any non-square `spawnBounds` or other `Rectangle3D` JSON is affected.)

## Fix

Omit Z only for the exact sentinel `Read` produces (`z1 == -128 && z2 == 127`); write it for anything else:

```csharp
var writeZ = value.Start.Z != sbyte.MinValue || value.End.Z != sbyte.MaxValue;
```

`Read` is unchanged, so existing `regions.json` rectangles that omit Z keep loading identically.

## Tests

New `Rectangle3DConverterTests`: round-trips the homeRange depth-256 case, the depth-255 sentinel, and ordinary/edge z values; asserts the sentinel omits Z while homeRange bounds write it. Server.Tests 717/717, UOContent.Tests 487/487.

### DIFF
--- a/Projects/Server/Json/Converters/Rectangle3DConverter.cs
+++ b/Projects/Server/Json/Converters/Rectangle3DConverter.cs
@@ -175,9 +175,6 @@ public class Rectangle3DConverter : JsonConverter<Rectangle3D>
 
     public override void Write(Utf8JsonWriter writer, Rectangle3D value, JsonSerializerOptions options)
     {
-        // Omit z only for the exact full-range sentinel DeserializeObj reconstructs when z is absent
-        // (z1 == -128, z2 == 127). Any other z must be written, otherwise it round-trips to that
-        // sentinel and is corrupted (e.g. a homeRange-style z1=-128/z2=128 would lose a z-level).
         var writeZ = value.Start.Z != sbyte.MinValue || value.End.Z != sbyte.MaxValue;
 
         writer.WriteStartObject();
--- a/Projects/Server/Json/Converters/Rectangle3DConverter.cs
+++ b/Projects/Server/Json/Converters/Rectangle3DConverter.cs
@@ -175,7 +175,10 @@ public class Rectangle3DConverter : JsonConverter<Rectangle3D>
 
     public override void Write(Utf8JsonWriter writer, Rectangle3D value, JsonSerializerOptions options)
     {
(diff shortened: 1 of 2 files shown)


Projects/Server/Json/Converters/Rectangle3DConverter.cs
```diff
@@ -175,7 +175,10 @@ public override Rectangle3D Read(ref Utf8JsonReader reader, Type typeToConvert,
 
     public override void Write(Utf8JsonWriter writer, Rectangle3D value, JsonSerializerOptions options)
     {
-        var writeZ = value.Start.Z is > sbyte.MinValue and < sbyte.MaxValue || value.End.Z is > sbyte.MinValue and < sbyte.MaxValue;
+        // Omit z only for the exact full-range sentinel DeserializeObj reconstructs when z is absent
+        // (z1 == -128, z2 == 127). Any other z must be written, otherwise it round-trips to that
+        // sentinel and is corrupted (e.g. a homeRange-style z1=-128/z2=128 would lose a z-level).
+        var writeZ = value.Start.Z != sbyte.MinValue || value.End.Z != sbyte.MaxValue;
```
kamronbatman:
```suggestion
        var writeZ = value.Start.Z != sbyte.MinValue || value.End.Z != sbyte.MaxValue;
```

PR #2554, opened 2026-07-29
fix(opl): don't tear down a property list that is mid-interpolation

kamronbatman:
## The bug

Any property getter reached from `GetProperties` that calls `InvalidateProperties` aborts the whole tooltip build:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'array')
   at Server.ObjectPropertyList.AppendFormatted(String value)
   at Server.Mobiles.PlayerMobile.GetProperties(IPropertyList list)
```

`InvalidateProperties` rebuilds in place:

```csharp
m_PropertyList.Reset();
InitializePropertyList(m_PropertyList);   // GetProperties runs with m_PropertyList set
```

`Reset()` calls `Dispose()`, which returns the interpolation scratch buffer to `STArrayPool` and nulls `_arrayToReturnToPool`. Re-entering that while a `$"..."` handler is still appending pulls the buffer out from under the in-flight handler, so the next `Append*` does `_arrayToReturnToPool.AsSpan(_pos..)` on a null array — `ArgumentNullException`, parameter `"array"`.

The first-build path happens to survive (`PropertyList` uses `??=`, so `m_PropertyList` is still null and the nested call recurses into a throwaway list instead), which is why this only shows up on rebuilds.

## The trigger in content

Factions `PlayerState.Rank` is exactly such a getter — it called `Invalidate()` from inside its own getter. `m_InvalidateRank` defaults to `true`, so the first tooltip rebuild for any faction member on the faction facet after a load threw out of `PlayerMobile.GetProperties`:

```csharp
list.Add(1060776, $"{pl.Rank.Title}\t{faction.Definition.PropName}"); // ~1_val~, ~2_val~
```

## The fix

Three layers, root cause first:

1. **`PlayerState.Rank` is side-effect free.** Invalidation moves to the `RankIndex` setter, where the rank is actually dirtied. The getter also now always resolves a rank: ranks are ordered by `Required` descending with a final `Required = 0`, so a *negative* percent (`RankIndex` out of sync with `ZeroRankOffset`) matched nothing and left `m_Rank` null **and** `m_InvalidateRank` true — which NREs on `Rank.Title` and re-invalidates on every subsequent build. It also no longer divides by a zero `ZeroRankOffset`.
2. **`Mobile` and `Item` defer a nested `InvalidateProperties`** raised during a build and push the delta once the build completes. The list under construction already observes the new state, so nothing is lost — and no other side-effecting getter anywhere in content can reach the crash.
3. **`ObjectPropertyList` re-rents its scratch buffer** instead of spanning a null array, so a stray `Reset()` from any caller degrades gracefully rather than aborting `GetProperties`.

## Tests

Adds `ObjectPropertyListReentrancyTests` covering `Reset()` and `Dispose()` re-entered mid-hole. Both are red before the change with the exact production exception.

790/790 `Server.Tests` and 603/603 `UOContent.Tests` pass.

## Note, not addressed here

`~ObjectPropertyList()` returns the rented array to `STArrayPool<char>.Shared` from the **finalizer thread**, and that pool is single-threaded by design. Left alone as a separate concern.

### DIFF
--- a/Projects/Server.Tests/Tests/PropertyList/ObjectPropertyListReentrancyTests.cs
+++ b/Projects/Server.Tests/Tests/PropertyList/ObjectPropertyListReentrancyTests.cs
@@ -1,0 +1,66 @@
+using System;
+using Xunit;
+
+namespace Server.Tests;
+
+/// <summary>
+/// The interpolation scratch buffer is rented by the compiler-generated handler ctor
+/// (InitializeInterpolation) and returned by the closing Add. Reset()/Dispose() also return it, so
+/// if either runs while a hole is still being evaluated the buffer is nulled underneath the
+/// in-flight handler and the next Append* spans a null array:
+///
+///   System.ArgumentNullException: Value cannot be null. (Parameter 'array')
+///     at Server.ObjectPropertyList.AppendStringDirect(String value)
+///
+/// This is reachable in practice. Mobile.InvalidateProperties rebuilds in place:
+///
+///   m_PropertyList.Reset();
+///   InitializePropertyList(m_PropertyList);   // <- GetProperties runs with m_PropertyList set
+///
+/// so any property getter with an InvalidateProperties side effect that is read from inside
+/// GetProperties re-enters, calls Reset() on the list being written, and kills the whole tooltip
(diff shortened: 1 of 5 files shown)


kamronbatman:
REMOVED FROM `main`

PR #2559, opened 2026-08-05
perf(core): sleep the event loop when idle

kamronbatman:
## Problem

`RunEventLoop` span through its body regardless of whether there was anything to do — ~10% of a desktop core for an empty shard, and ~70% of a core on a 3 vCPU VPS. A process that never idles is exactly what burstable vCPU plans throttle, which is how this surfaced: lag spikes that went away when the operator bought more cores. The spin also denied the GC its natural pause points, so memory climbed until a world save forced a collection — alarming in task manager, harmless in practice, and a recurring source of "is my server leaking?" reports.

## Result

Windows desktop, real world of **190,728 items / 33,158 mobiles**, no players, saves and prebake off, three consecutive runs:

| | Legacy spin | Idle sleeping |
|---|---|---|
| **CPU** | 10.42 – 10.50% of one core | **0.78 – 1.00%** |
| **Tick lag** (peak/15s) | 4–10 ms | 5–11 ms |

**~10× less CPU with tick lag unchanged** — the CPU came free rather than being traded for latency. Slower hosts gain proportionally more. Spin mode (`server.eventLoopIdleWaitMs=0`) independently gained **7× the iterations per core** (1.19M → 8.3M cycles/sec) from the ring's AcceptEx rework.

## How

The loop blocks in `NetState.WaitForCompletion` whenever every queue it drains is empty (all the drains are bounded, so leftovers keep it awake). Receive completions, new connections, and cross-thread `LoopContext.Post` (via the ring's sticky `Wake()`) are all in the wait set, so sleeping adds no latency to any of them. Only timer-driven logic sees wheel lag, bounded by the idle wait.

**Health is measured at the only place sleeping can cause harm.** A sleep is bounded by the time to the next wheel turn, so a correctly honoured sleep can never miss a deadline — the only failure mode is the host returning the wait late. That overshoot is measured on every sleep (one extra timestamp read; production's entire accounting cost), and an escalating backoff suspends sleeping when it persists. By construction, server work — saves, heavy staff commands, deep timer callbacks — cannot trip it, so the warning means exactly one thing: *the host is not scheduling the process promptly*, with two known remedies (dedicated CPU, or `=0`). Hosts with no high-resolution wait mechanism at all are detected once at startup and spin instead.

**CPS is removed.** `Core.CyclesPerSecond`/`AverageCPS` measured nothing actionable before and became actively misleading once the loop sleeps (the rate is set by the sleep, not by shard health). The admin gump's Performance page now shows the verdict instead: `Healthy` / `Sleep suspended (host)` / `Spinning (configured)`.

## Configuration

| Setting | Default | Meaning |
|---|---|---|
| `server.eventLoopIdleWaitMs` | `2` | Longest idle block. Measured across 1/2/4/8 ms, 2 is where the trade stops being free. `0` = never sleep: ~98% of a core, zero scheduling overhead — for large shards on dedicated CPU. |
| `server.lateWakeThreshold` | `1` | Idle waits the host may return a full tick late, per second, before sleeping backs off. Raise for jittery hosts; very high disables the backoff. |

## Diagnostics (compiled out by default)

`dotnet build -p:EventLoopProfiling=true` compiles in `EventLoopProfiler` — every hook is `[Conditional("EVENT_LOOP_PROFILING")]`, so normal builds contain zero profiling IL. The profiling build decomposes each second of wall time into **work (per loop phase) / sleep / GC pause / stolen residual**, keeps ~15 minutes of history in a ring buffer, and the `[LoopStats` command prints the last minute and dumps the full history to CSV. `dev-docs/debugging-event-loop.md` is the diagnosis guide (for humans and AI): what production already tells you, when to flip the profiling build, the signature table for host-steal vs deep-processing vs GC vs wake bugs, why dotnet-trace comes last, and the GC/RAM "leak" misconception.

## Verification

- 815 Server.Tests green; both build configurations compile.
- Docker echo harness green on epoll and io_uring (ping-pong mode); kqueue verified manually on an M1 Max.
- A/B measurements and per-change numbers: `measure/event-loop` branch.

## Notes

The full measurement harness and vendored ring sources used to develop this live on the [`measure/event-loop`](https://github.com/modernuo/ModernUO/tree/measure/event-loop) branch, kept for future loop work.

### DIFF
--- a/Projects/Server/Main.cs
+++ b/Projects/Server/Main.cs
@@ -617,8 +617,8 @@ public static class Core
         if (_eventLoopIdleWaitMs > 0 && NetState.Ring?.SupportsHighResolutionWait == false)
         {
             logger.Error(
-                "This host cannot honour short waits (no high-resolution timer, and raising the system timer " +
-                "resolution failed). Idle sleeping is disabled -- the loop will spin instead, using a full core."
+                "This host cannot honor short waits (no high-resolution timer, and raising the system timer " +
+                "resolution failed). Idle sleeping is disabled. The loop will spin instead, using a full core."
             );
 
             _eventLoopIdleWaitMs = 0;
--- a/Projects/Server/Main.cs
+++ b/Projects/Server/Main.cs
@@ -39,10 +39,142 @@ public static class Core
 {
     private static readonly ILogger logger = LogFactory.GetLogger(typeof(Core));
 
-    private static bool _performProcessKill;
+    // Written from other threads (Kill, RequestSnapshot) and read by the event loop. Volatile
+    // because the loop now genuinely blocks between reads rather than spinning past them.
+    private static volatile bool _performProcessKill;
     private static bool _restartOnKill;
-    private static bool _performSnapshot;
+    private static volatile bool _performSnapshot;
     private static string _snapshotPath;
+
+    // A backstop, not a latency control: the wheel's tick rate bounds the sleep, so this only
+    // limits the damage if a wake signal is ever missed. Measured across 1/2/4/8ms, 2 is optimal.
+    private static int _eventLoopIdleWaitMs = 2;
+
+    /// <summary>
+    /// Longest the loop will block while idle, in milliseconds. 0 disables idle sleeping,
+    /// leaving the loop to spin; the adaptive backoff does the same thing temporarily when the
+    /// host keeps returning waits late.
+    /// </summary>
+    public static int EventLoopIdleWaitMs => _eventLoopIdleWaitMs;
+
+    /// <summary>
+    /// Whether idle sleeping is currently suspended because the host returned waits late.
+    /// </summary>
+    /// <remarks>
+    /// Compared by subtraction, never directly: tick counts can start enormous and wrap.
+    /// See dev-docs/tick-counts.md.
+    /// </remarks>
+    public static bool IdleSleepSuspended => _tickCount - _idleSleepSuspendedUntil < 0;
+
+    private const long HealthSampleIntervalMs = 1000;
+
+    // Backoff escalates by doubling: a fixed suspension oscillates forever on a persistently bad
+    // host, while doubling converges on "stop sleeping" within minutes yet still recovers from a
+    // transient problem.
+    private const long BackoffBaseMs = 5000;
+    private const long BackoffMaxMs = 120_000;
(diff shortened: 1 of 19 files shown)


kamronbatman:
## Repurposed as the measurement branch — do not merge

This branch now **vendors IORingGroup as source** under `Projects/IORingGroup/` and no longer references the package. Clone and `dotnet build`; there is nothing to restore.

**Why:** coupling the repos through NuGet means every design iteration burns a published version, and each published version needs its own measurement pass to know whether it was the right one. A wrong guess is then stuck on the feed permanently. Vendoring decouples them so the measurements can decide the design before anything is published.

The vendored sources are byte-identical to [modernuo/IORingGroup#11](https://github.com/modernuo/IORingGroup/pull/11) (verified by hash across all 18 files), so what is measured here is exactly what would ship.

**Sequencing from here:**
1. Gather measurements on real hardware — the gaps are a populated shard (500+ concurrent) and kqueue/macOS.
2. Let those results guide the final design.
3. Publish the settled IORingGroup version.
4. Build the real ModernUO PR: restore the `PackageReference`, delete `Projects/IORingGroup/`, drop the throw-away bits.

`measure/event-loop-2026-08-04` points at the same commit and is the stable handle for community runs, since this branch will keep moving.

**IORingGroup#11 should not be merged or published until step 2 is done.**

Projects/Server/Main.cs
```diff
@@ -469,34 +610,63 @@ You should have received a copy of the GNU General Public License
...
         EventSink.InvokeServerStarted();
+
+        // Without a high-resolution wait a 2ms request quantises to 15.625ms and the loop would
+        // quietly run a tick behind; spinning is the lesser evil and must not be silent. Only
+        // fires when both the ring's high-res timer and its timeBeginPeriod fallback failed.
+        if (_eventLoopIdleWaitMs > 0 && NetState.Ring?.SupportsHighResolutionWait == false)
+        {
+            logger.Error(
+                "This host cannot honour short waits (no high-resolution timer, and raising the system timer " +
+                "resolution failed). Idle sleeping is disabled -- the loop will spin instead, using a full core."
```
kamronbatman:
```suggestion
                "This host cannot honor short waits (no high-resolution timer, and raising the system timer " +
                "resolution failed). Idle sleeping is disabled. The loop will spin instead, using a full core."
```

PR #2563, opened 2026-08-08
fix: Warn when sending empty gumps

### LINKED ISSUE #1989: Sending invisible (empty) gumps
Sending an empty (invisible gump) should produce a warning in debug mode.

Guflly:
Fixes #1989.

Warns in debug builds when a legacy, dynamic, or static gump has no visible elements. Structural entries such as pages and flags do not suppress the warning.

Tests cover empty and visible gumps across all three implementations.

### DIFF
--- a/Projects/UOContent/Gumps/Base/BaseGump.cs
+++ b/Projects/UOContent/Gumps/Base/BaseGump.cs
@@ -24,9 +24,7 @@ namespace Server.Gumps;
 public abstract class BaseGump
 {
     private static readonly byte[] _packetBuffer = GC.AllocateUninitializedArray<byte>(0x10000);
-#if DEBUG
     private static readonly ILogger _logger = LogFactory.GetLogger(typeof(BaseGump));
-#endif
     private static Serial nextSerial = (Serial)1;
 
     public int TypeID { get; protected set; }
@@ -61,12 +59,10 @@ public abstract class BaseGump
         var writer = new SpanWriter(_packetBuffer);
         Compile(ref writer);
 
-#if DEBUG
         if (!HasVisualElements)
         {
             _logger.Warning("Sending empty gump {GumpType}", GetType().FullName);
         }
-#endif
 
         ns.Send(writer.Span);
 
--- a/Projects/UOContent/Gumps/Base/BaseGump.cs
+++ b/Projects/UOContent/Gumps/Base/BaseGump.cs
@@ -31,7 +31,7 @@ public abstract class BaseGump
 
     public int TypeID { get; protected set; }
     public Serial Serial { get; protected set; }
-    internal bool HasVisualElements { get; set; }
+    protected bool HasVisualElements { get; set; }
 
     public abstract int Switches { get; }
     public abstract int TextEntries { get; }
--- a/Projects/UOContent/Gumps/Base/BaseGump.cs
+++ b/Projects/UOContent/Gumps/Base/BaseGump.cs
@@ -13,6 +13,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>. *
  *************************************************************************/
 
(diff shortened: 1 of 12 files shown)


Projects/UOContent/Gumps/Base/GumpLayoutBuilder.cs
```diff
@@ -30,6 +30,7 @@ public ref struct GumpLayoutBuilder
     internal GumpFlags _flags;
     internal int _switches;
     internal int _textEntries;
+    internal int _visualElements;
```
kamronbatman: This could probably be a boolean.

Guflly: Updated this to a bool.

Projects/UOContent/Gumps/Base/StaticGump.cs
```diff
@@ -30,6 +30,7 @@ public abstract class StaticGump<TSelf> : BaseGump where TSelf : StaticGump<TSel
     private static byte[] _compressedStringsData;
 
     private static bool _hasDynamicStrings;
+    private static bool _hasVisualElements;
```
kamronbatman: StaticGump is a BaseGump, so we should have access to `HasVisualElements` and therefore do not need to redeclare it.

Projects/UOContent/Gumps/Base/BaseGump.cs
```diff
@@ -23,10 +24,14 @@ namespace Server.Gumps;
...
 {
     private static readonly byte[] _packetBuffer = GC.AllocateUninitializedArray<byte>(0x10000);
+#if DEBUG
+    private static readonly ILogger _logger = LogFactory.GetLogger(typeof(BaseGump));
+#endif
     private static Serial nextSerial = (Serial)1;
 
     public int TypeID { get; protected set; }
     public Serial Serial { get; protected set; }
+    internal bool HasVisualElements { get; set; }
```
kamronbatman: `internal` should probably be `protected`?

kamronbatman:
Thanks for taking care of this!

Projects/UOContent/Gumps/Base/StaticGump.cs
Guflly: This needs a static cache for later instances that reuse the compressed layout. I renamed it to `_cachedHasVisualElements` and added coverage for the cached instance.

Projects/UOContent/Gumps/Base/BaseGump.cs
Guflly: Changed this to `protected`.

Projects/UOContent/Gumps/Base/StaticGump.cs
kamronbatman: Yeah interesting, I just realized it's static. Ok that makes sense!

Projects/UOContent/Gumps/Base/BaseGump.cs
```diff
@@ -56,6 +61,13 @@ public virtual void SendTo(NetState ns)
         var writer = new SpanWriter(_packetBuffer);
         Compile(ref writer);
 
+#if DEBUG
+        if (!HasVisualElements)
+        {
+            _logger.Warning("Sending empty gump {GumpType}", GetType().FullName);
+        }
+#endif
```
kamronbatman: I am on the fence about this. I think either we make this available all the time and annoy the admin, or the entire `HasVisualElements` machinery is behind `#if DEBUG` so we don't bother computing/wasting memory for no reason.

I am leaning toward annoying the admins because it is an actual client leak.

Guflly: Made the warning run in all builds.

PR #2576, opened 2026-08-14
fix: stop stack merges and splits from laundering PlayerConstructed

kamronbatman:
## Why

`PlayerConstructed` is per-instance provenance, and #2574 put it on every crafted item — including potions, arrows and other stackables. Stack operations were written when no item carried provenance of any kind, so they treated two piles of the same graphic as interchangeable.

**Merging** keeps the receiving stack's value. Dropping bought potions onto a crafted stack made the whole pile count as crafted; the reverse order erased it. Which one happened was decided by drag direction alone.

**Splitting** rebuilds one half in `Mobile.LiftItemDupe`, which copies a fixed list of fields rather than going through `Dupe`/`CopyProperties`. `PlayerConstructed` was not on that list, so dragging part of a pile off stripped the new half. Worth calling out: `[IgnoreDupe]` does **not** govern this path — it only applies to `Dupe()`. Reasoning "the field isn't `[IgnoreDupe]`, so it copies" is wrong here.

## Changes

- `Item.CanStackWith` compares `PlayerConstructed`, so crafted and non-crafted never merge into one indistinguishable pile.
- `Mobile.LiftItemDupe` copies `PlayerConstructed` onto the remainder, so a split cannot produce halves that disagree about what they are.

Refusing to merge is the whole fix. A stack has nowhere to record provenance, so the only coherent behaviour is to keep the two piles apart rather than pick a winner.

## What this deliberately does not do

Paths that genuinely **virtualize** an item — pouring from a `PotionKeg`, for one — rebuild it without the flag, and the result is simply treated as not crafted. That is accepted rather than worked around; the alternative is threading provenance through every count-based container, which buys little. The keg stores a `Held` int rather than a stack, so nothing there depends on merging and nothing breaks.

`CommodityDeed` is unaffected — it holds the real `Commodity` item rather than a count, so the flag rides along.

## Player-visible effect

Crafted potions and arrows will no longer stack with bought or looted ones. That is the intended invariant, and it is the reason the flag can be trusted at all.

## Tests

7 new tests in `Server.Tests`: both merge directions, the matching-provenance case, split copying, and the split/re-merge round trip.

`Server.Tests` **822 passing**, `UOContent.Tests` **701 passing**, build clean with 0 warnings.

### DIFF
--- a/Projects/Server/Items/Item.cs
+++ b/Projects/Server/Items/Item.cs
@@ -2344,12 +2344,16 @@ public partial class Item : IHued, IComparable<Item>, ISpawnable, IObjectPropert
         UpdateDecayRegistration();
     }
 
+    // PlayerConstructed is part of stack identity. A merge keeps the receiving stack's value, so
+    // without this term the outcome depends on which side was dropped onto which: one order
+    // launders the flag onto items that never earned it, the other erases it from items that did.
     public virtual bool CanStackWith(Item dropped) =>
         dropped.Stackable && Stackable &&
         dropped.GetType() == GetType() &&
         dropped.ItemID == ItemID &&
         dropped.Hue == Hue &&
         dropped.Name == Name &&
+        dropped.PlayerConstructed == PlayerConstructed &&
         dropped.Amount + Amount <= 60000 &&
         dropped != this;
 
--- a/Projects/Server/Items/Item.cs
+++ b/Projects/Server/Items/Item.cs
@@ -2344,9 +2344,6 @@ public partial class Item : IHued, IComparable<Item>, ISpawnable, IObjectPropert
         UpdateDecayRegistration();
     }
 
-    // PlayerConstructed is part of stack identity. A merge keeps the receiving stack's value, so
-    // without this term the outcome depends on which side was dropped onto which: one order
-    // launders the flag onto items that never earned it, the other erases it from items that did.
     public virtual bool CanStackWith(Item dropped) =>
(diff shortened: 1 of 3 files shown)


Projects/Server/Mobiles/Mobile.cs
```diff
@@ -5248,6 +5248,12 @@ public static T LiftItemDupe<T>(T oldItem, int amount) where T : Item
         item.Name = oldItem.Name;
         item.Weight = oldItem.Weight;
 
+        // Splitting a stack must not change what either half is. This copies a fixed list of
+        // fields rather than going through Dupe/CopyProperties, so anything omitted here is
+        // silently dropped on the new half -- for PlayerConstructed that would mean dragging
+        // part of a pile off is enough to strip its provenance.
+        item.PlayerConstructed = oldItem.PlayerConstructed;
+
```
kamronbatman:
```suggestion
        item.PlayerConstructed = oldItem.PlayerConstructed;
```

Projects/Server/Items/Item.cs
```diff
@@ -2344,12 +2344,16 @@ public void SetLastMoved()
         UpdateDecayRegistration();
     }
 
+    // PlayerConstructed is part of stack identity. A merge keeps the receiving stack's value, so
+    // without this term the outcome depends on which side was dropped onto which: one order
+    // launders the flag onto items that never earned it, the other erases it from items that did.
```
kamronbatman:
```suggestion
```

PR #2592, opened 2026-08-23
refactor: BaseCreature to the SerializationGenerator with SaveFlag elision

kamronbatman:
Converts BaseCreature's hand-written v22 serialization to the SerializationGenerator (v23), aimed at save freeze time and disk for worlds with 500k–1M creatures.

### The win

~40 fields sit behind `[SaveFlag]` (a single `ulong` flags word). A creature matching its defaults — most of a live world — writes a **13-byte tail** (4 version + 8 flags + 1 default-AI) instead of ~236 bytes, and the writer's work becomes ~40 cheap predicates plus one flags write instead of ~50 unconditional writes. Written ints are `[EncodedInt]`.

### Table speeds are not persisted

The speed flags and their load defaults both resolve through `GetSpeeds`/`GetMoveSpeeds`: a creature whose speeds match its npc-speeds bucket writes nothing and re-reads the table at load. `CurrentSpeed` writes only when it differs from `PassiveSpeed`. Consequences:

- npc-speeds.json edits now reach existing unmodified spawns on restart (the table is authoritative for untouched creatures).
- Former paragons elide fully (UnConvert's snap-to-table); live paragons legitimately persist their scaled speeds.

### Structure

- **Stateful `SpeedClass` + cached speed entry**: the four per-save table checks (and elided-load defaults) read a `SpeedClassEntry` cached on the creature instead of walking the dictionaries each time. To keep the cache impossible to hold wrong, `SpeedClass` is no longer virtual: a type's constant bucket is `virtual DefaultSpeedClass`, and runtime state changes (boss enrage, `[props`) *assign* `SpeedClass` — the setter invalidates the cache, applies the new bucket's think/move speeds (preserving active/passive mode), and serializes only when it differs from the type default, so a changed bucket survives a save while bucket-matching speeds still elide. `GetSpeeds`/`GetMoveSpeeds` remain the override point for bypassing the table, with no cache in that path.
- **The four speeds are one block, and `None` means custom**: either all of active/passive think + move conform to the bucket (elided as a set) or the creature is custom and all four serialize — partial conformance cannot exist on the wire. The old `None` → type list → Medium chain was construction-time defaulting, so the constructor now resolves it once into `SpeedClass` itself: at runtime a concrete bucket means table-backed, `None` means the creature's own speeds are authoritative. Tuning any speed flips the bucket to `None` (the label never lies), assigning a real bucket un-customs, runtime entry resolution is a single level lookup, and legacy loads demote nonconforming creatures (`SetSpeed` vendors) to `None` honestly.

- **Old saves (v0–22)** load through the retained legacy `Deserialize(reader, version)`, now assigning raw fields. Shared post-load fixups (stat timers, `ChangeAIType`, followers, animate-dead registration, unsummon timer, abandoned-pet fallback) moved to a single `[AfterDeserialization]` hook serving both paths.
- **Delete countdown** is a `[DeserializeTimer]` field (anchored — downtime does not consume it, matching the old remaining-`TimeSpan` semantics). Stabled/controlled pets never persist one; the 3-day `LastOwner` fallback lives in the hook. `SummonEnd` stays `[AnchoredDateTime]`, written only while summoned.
- **Side-effect setters became generated-field hooks**: `Team`, `Controlled`, `Summoned`, resistance seeds (`UpdateResistances`), `CurrentSpeed` (AI timer notify), `Loyalty` (clamp via `allowFieldChange`).
- **Five properties stay hand-written** as `[SerializableProperty]` with `useField:` — `ControlMaster`/`SummonMaster` (follower bookkeeping brackets the assignment), `ControlOrder` (order logic must run on equal re-assignment for pet commands), `Tamable`/`IsParagon` (custom getter / conversion must not run at load). `ActiveMoveSpeed`/`PassiveMoveSpeed` collapsed to plain `[SerializableField]`s (no longer virtual): they read the raw override (`0` = inheriting) and `CurrentMoveSpeed` carries the inherit resolution.
- **`CreatureDeathEvent`/`CreatureDeletedEvent` moved to a `CreatureEvents` host class**: the events generator and the serialization generator each emit a `[GeneratedCode]` partial for the declaring type, and the attribute forbids duplicates (CS0579) — BaseCreature is the first class using both. Can fold back if this is ever fixed generator-side.
- **One serialized master reference**: `ControlMaster` and `SummonMaster`, when both set, are always the same mobile (`Summon` assigns both; every management flow set them in lockstep) — they differ only in presence (uncontrolled summons vs pets). One `_master` ref serializes and fans back out through the `Controlled`/`Summoned` flags at load; `SetControlMaster` now maintains the lockstep itself, deleting six hand-rolled copies of that boilerplate.
- **Missing npc-speeds.json degrades gracefully**: a creature constructed without speeds logs debug and defaults to Medium (0.25/0.5) instead of spawning a 0-delay AI timer.
- `m_` fields renamed to `_camelCase`; `MoveSpeedMod`-era vestiges stay gone.

### Testing

- New `BaseCreatureSerializationTests`: default and fully-populated round trips with exact byte-consumption asserts; back-to-back saves byte-identical (anchored timer keeps idle saves stable); bucket assignment and partial-tuning-flips-to-custom round trips. (A byte-authentic fossilized v22 stream was validated through the legacy path during development, then dropped to avoid test bloat.)
- Full suite passes (1561). `Server.Mobiles.BaseCreature.v23.json` generated and committed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Projects/UOContent.Tests/Tests/Mobiles/AI/MoveSpeedTests.cs
+++ b/Projects/UOContent.Tests/Tests/Mobiles/AI/MoveSpeedTests.cs
@@ -57,8 +57,9 @@ public class MoveSpeedTests : IDisposable
     {
         var bc = NewCreature();
 
-        Assert.Equal(0.3, bc.ActiveMoveSpeed);
-        Assert.Equal(0.6, bc.PassiveMoveSpeed);
+        // 0 = no override; the resolved pace comes from CurrentMoveSpeed.
+        Assert.Equal(0, bc.ActiveMoveSpeed);
+        Assert.Equal(0, bc.PassiveMoveSpeed);
         Assert.Equal(bc.CurrentSpeed, bc.CurrentMoveSpeed);
     }
 
@@ -96,8 +97,8 @@ public class MoveSpeedTests : IDisposable
 
         bc.SetSpeed(0.2, 0.4);
 
-        Assert.Equal(0.2, bc.ActiveMoveSpeed);
-        Assert.Equal(0.4, bc.PassiveMoveSpeed);
+        Assert.Equal(0, bc.ActiveMoveSpeed);
+        Assert.Equal(0, bc.PassiveMoveSpeed);
     }
 
     [Fact]
@@ -108,8 +109,10 @@ public class MoveSpeedTests : IDisposable
 
         bc.ActiveMoveSpeed = 0;
 
-        Assert.Equal(0.3, bc.ActiveMoveSpeed);  // inheriting again
+        Assert.Equal(0, bc.ActiveMoveSpeed);    // inheriting again
         Assert.Equal(0.9, bc.PassiveMoveSpeed); // other override untouched
+        bc.SetCurrentSpeedToActive();
+        Assert.Equal(0.3, bc.CurrentMoveSpeed); // resolves to the think clock
     }
 
     [Fact]
@@ -121,7 +124,7 @@ public class MoveSpeedTests : IDisposable
         bc.ScaleMoveSpeed(1.0 / 1.2);
 
         Assert.Equal(0.5, bc.ActiveMoveSpeed);
-        Assert.Equal(bc.PassiveSpeed, bc.PassiveMoveSpeed); // still inheriting, not 0 * scalar
+        Assert.Equal(0, bc.PassiveMoveSpeed); // still inheriting, not 0 * scalar
     }
 
     [Fact]
@@ -185,8 +188,8 @@ public class MoveSpeedTests : IDisposable
 
         bc.MigrateMoveSpeeds();
 
-        Assert.Equal(0.35, bc.ActiveMoveSpeed);
-        Assert.Equal(0.6, bc.PassiveMoveSpeed);
+        Assert.Equal(0, bc.ActiveMoveSpeed); // still inheriting the (tuned) think clock
+        Assert.Equal(0, bc.PassiveMoveSpeed);
     }
 
     [Theory]
@@ -213,7 +216,7 @@ public class MoveSpeedTests : IDisposable
 
         // The v22 tail is the last block; exact consumption catches any offset mistake.
         Assert.Equal(buffer.Length, reader.Position);
-        Assert.Equal(overridden ? 0.45 : 0.3, copy.ActiveMoveSpeed);
-        Assert.Equal(overridden ? 0.9 : 0.6, copy.PassiveMoveSpeed);
+        Assert.Equal(overridden ? 0.45 : 0, copy.ActiveMoveSpeed);
+        Assert.Equal(overridden ? 0.9 : 0, copy.PassiveMoveSpeed);
     }
 }
--- a/Projects/UOContent.Tests/Tests/Mobiles/BaseCreatureSerializationTests.cs
+++ b/Projects/UOContent.Tests/Tests/Mobiles/BaseCreatureSerializationTests.cs
@@ -85,7 +85,7 @@ public class BaseCreatureSerializationTests : IDisposable
         Assert.Equal(0.3, copy.ActiveSpeed);
         Assert.Equal(0.6, copy.PassiveSpeed);
         Assert.Equal(0.6, copy.CurrentSpeed);
-        Assert.Equal(0.6, copy.ActiveMoveSpeed);   // pulled from the table, not the wire
+        Assert.Equal(0.6, copy.ActiveMoveSpeed);   // class None (no table in tests): restored from the wire
         Assert.Equal(1.2, copy.PassiveMoveSpeed);
         Assert.Equal(100, copy.PhysicalDamage);
         Assert.Equal(BaseCreature.MaxLoyalty, copy.Loyalty);
@@ -238,7 +238,7 @@ public class BaseCreatureSerializationTests : IDisposable
 
         bc.ActiveSpeed = 0.25; // one tuned value customizes the whole block
 
-        Assert.Equal(SpeedLevel.Custom, bc.SpeedClass); // the bucket label never lies
+        Assert.Equal(SpeedLevel.None, bc.SpeedClass); // the bucket label never lies
 
         var writer = new BufferWriter(true);
         bc.Serialize(writer);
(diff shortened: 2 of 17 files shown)


kamronbatman:
## Audit: ControlMaster vs SummonMaster, for the master collapse

The pure generator conversion moved to #2611 (Delta Saves #7, phase 3). This PR now carries two behaviour changes on top of it: SpeedClass and the single-`Master` collapse. An inventory of every read and write of the two references was done to decide whether the collapse is safe. Summary:

**Writes.** Every writer sets them together or sets exactly one, gated by the flags:
- `Summon(controlled: true)` (every `SpellHelper.Summon` spell, familiars, steeds, talismans): both = caster.
- `Summon(controlled: false)` (Blade Spirits, Energy Vortex, Animated Weapon, Nature's Fury, Vengeful Spirit, Animate Dead), Mirror Image clones, Summon Undead counter summons: `SummonMaster` only, `Summoned = true`, never `Controlled`.
- `SetControlMaster(x)` (taming, hires, escorts, vendors, Clockwork Assembly, Horse Breeder): `ControlMaster` only. The claim/transfer/obey paths then copy it into `SummonMaster` only `if (Summoned)`.
- `EnragedCreature`: `SummonMaster` only with `Summoned = false` (MeerMage's summons, deliberately not dispellable and not a pet). `OnThink` deletes it when the master is gone.
- `BaseTalisman.Summon`: controlled summon, then `Summoned = false` so it behaves as a pet (kill awards, `(tame)`, stable-able) with an unsummon timer.
- `SetControlMaster(null)` (release order, loyalty 0, escort/hire abandon, jail): clears `ControlMaster` only; `SummonMaster` is kept.

**Reads.** `GetMaster()` is flag-gated (`Controlled ? CM : Summoned ? SM : null`). Eleven sites use `CM ?? SM` (follower bookkeeping, pack instinct, familiars, guild lookup). Every ungated `ControlMaster` read only ever sees a value when `Controlled`, and every ungated `SummonMaster` read only when `Summoned`, with one exception: `EnragedCreatures.cs` (reads `SummonMaster` with `Summoned` false, and its line 121 presence test `bc.Controlled || bc.SummonMaster != null` on other creatures).

**The only state with two different non-null masters** is reachable but accidental: a talisman summon transferred to another player (`Summoned` is false, so the transfer does not update `SummonMaster`) ends with `ControlMaster = to`, `SummonMaster = from` for its remaining minutes. The only observable effect is that the original summoner's area spells exempt it. Nothing reads the disagreement on purpose.

**A released summon keeps its caster on purpose.** After "release" or loyalty 0 on a controlled summon, `ControlMaster` is null but `SummonMaster` stays: `GetMaster()` still credits the caster, `BaseAI.IsInvalidSummonTarget` still stops it attacking the caster, `Notoriety` still shields it from other players, the follower slot moves to `SummonMaster`. That is the one place the collapse must be careful.

**Recommendation for this PR:** one serialized `Master` plus the existing `Controlled`/`Summoned` flags is sound; no separate "dispellable" flag is needed (the `Summoned = false` users rely on all of `Summoned`'s side effects, not just dispel, and `IsDispellable` is already virtual). Rules the fold must implement, which the earlier `AfterDeserialization` fan-out did not:
1. `SetControlMaster(null)` clears `Controlled` and clears `Master` only when `!Summoned`.
2. `EnragedCreature` reads and writes `Master` directly (a `Summoned`-gated `SummonMaster` view would delete it on the first `OnThink`); line 121 becomes `bc.Controlled || bc.Master != null`.
3. Load migration: `Master = ControlMaster ?? SummonMaster`; when both are set and differ (the transfer case), keep `ControlMaster` and log.
4. `GetMaster()` keeps its meaning: `Controlled || Summoned ? Master : null`.

Sites that then become no-ops and can be deleted: the `if (Summoned) SummonMaster = x` copies in `AnimalTrainer`, `PlayerMobile` auto-claim, `BallOfSummoning`, `TransferItem`, `OnSpeech` (GM obey), and `HorseBreederGump`'s direct `Controlled/ControlMaster` writes (use `SetControlMaster`). Optional: the eleven `CM ?? SM` reads and `SpellHelper.GetGuildFor` collapse to `Master`.

Behaviour deltas, both arguably fixes: a transferred talisman summon no longer exempts its original summoner from area harm; a jail-stabled talisman summon no longer holds the summoner's follower slot.

kamronbatman:
## Correction to the audit above: a summon can never be released

The "released summon keeps its caster" paragraph is wrong, and with it the `!Summoned` clause in rule 1. Every path that could clear `ControlMaster` on a summon kills or deletes it first:

- Release order: `PetOrderHandlers.cs:252-256` does `if (Mobile.Summoned) { Mobile.Kill(); return; }` before `SetControlMaster(null)`; the think-side `DoOrderRelease` (`PetOrders.cs:479-481`) deletes when `DeleteOnRelease`, which is `=> _summoned` (`BaseCreature.cs:1256`). Summons are not bondable, so `Kill` ends in `Delete`. That is OSI behaviour (dismiss), not a bug.
- Loyalty drain: needs ten hours of continuous, commandable control; no controlled summon lives that long and is commandable (spell summons are minutes, ethics steeds one hour, necro familiars and talisman summons are `Commandable => false`). If it fired it would delete anyway.
- Transfer and stable refuse summons; jail dispels them; death, `Dispel`, `UnsummonTimer` all delete.

So `Controlled`, `Summoned` and the master are always consistent for every creature except `EnragedCreature` (and the transferred talisman summon, where the stale `SummonMaster` is harmless). Revised rules for the collapse:

1. `SetControlMaster(null)` clears `Master` unconditionally. Optional hardening: assert or dismiss when `Summoned`, so a future caller cannot recreate a split state.
2. `EnragedCreature` reads and writes `Master` directly; its line 121 presence test becomes `bc.Controlled || bc.Master != null`.
3. Load migration: `Master = ControlMaster ?? SummonMaster`; keep `ControlMaster` and log when both are set and differ.
4. `GetMaster()` keeps its meaning: `Controlled || Summoned ? Master : null`.

**Separate bug found (pre-existing, tame pets, not summons):** the loyalty drain at `BaseCreature.cs:5928-5941` calls `c.AIObject.DoOrderRelease()` directly instead of setting `ControlOrder = Release`, so it never reaches `HandleReleaseOrder` and `SetControlMaster(null)` (`PetOrderHandlers.cs:272`). A pet whose loyalty hits zero says it is better off without a master, gets the three-day delete timer and drops its pack, but stays `Controlled` with its `ControlMaster` and keeps holding the owner's follower slots, the opposite of the comment's intent. Worth its own fix outside this PR.

PR #2601, opened 2026-08-31
feat: event-driven target acquisition with a reaction-time gradient

kamronbatman:
Fixes walk-up aggro latency (up to a full 10 s of obliviousness) and hardens the reacquire gate so no state can silence acquisition, while turning `AcquireOnApproach` into the reaction-time knob for future per-creature intelligence tuning.

### Why

`AcquireFocusMob` re-armed the 10 s `ReacquireDelay` **before** scanning, success or failure. A creature that scanned an empty room was blind for 10 s to a player walking up — walk-up aggro latency was uniform in 0..10 s. Waking from sector sleep stacked the AI timer's 0–3 s construction stagger on top. And `NextReacquireTime` is not serialized: on hosts whose tick counter starts negative (GCP pass-through), the 0 default blocked **all** acquisition shard-wide after a restart until the counter crossed zero.

### What

**Event-driven reaction — `AcquireOnApproachDelay` (the intelligence gradient)**
- The paragon `AcquireOnApproach` bool becomes a `TimeSpan` on every creature: an enemy moving inside `AcquireOnApproachRange` (10 for all creatures — on-screen reactive aggro; the periodic scan keeps the wide `RangePerception` sweep) *clamps* the next scan to at most the delay. Repeated steps cannot shorten it further — one scan per delay period, not per step or think.
- `Zero` (paragons) also prods the AI timer: the ranked scan engages within a wheel turn — the old snap, minus the special-cased engage path. The target now comes from the normal FightMode ranking instead of whichever mobile happened to move, and the `Combatant == null` guard stops re-engage spam.
- The 2 s default reads as "took a beat to notice you"; larger values are dumber; `ReacquireDelay` alone is the oblivious floor. Mover checks are the approach logic's `IsEnemy` + `CanBeHarmful` (so pets count and hidden movers are excluded via `CanSee`), with `IsEnemy` first to cheaply reject same-team wild creatures wandering past. The check rides the `OnMovement` callback every step already pays for — no polling added.

**Gate correctness**
- Every scan re-arms the full `ReacquireDelay`, success or failure (classic semantics; reaction time is the approach path, not the poll).
- Self-healing by construction: a deadline further out than `ReacquireDelay` is an illegal state and reads as open — no wedged or wrapped value can silence acquisition beyond one delay period.
- `NextReacquireTime` is seeded from a live tick on deserialize (the GCP negative-tick blackout).

**AI timer wake**
- Activation (sector wake, spawn, resurrection) starts within a 0–256 ms spread instead of the 0–3 s construction stagger, which read as lag.
- The stagger's real job — keeping same-speed cohorts out of lock-step (the RunUO town artifact) — is now a zero-mean ±period/8 jitter on each **idle** think, so phases random-walk apart within seconds and can never re-lock. Instrumentation showed why a one-shot spread can't do this job: the timer wheel fires within ±1 ms, so with 10 creatures on a 500 ms period some pair collides on nearly the same phase ~75% of the time (birthday paradox) and then steps in the same loop iteration *forever*. Jitter is scoped to passive speed: engaged cadence stays exact, since pursuit timing anchors to real step times.

**Debug**
- The `AcquireFocusMob` scan message no longer re-arms the shared 5 s debug cooldown, which swallowed every AI's "I have detected X" transition line.

**API change** for custom scripts: `AcquireOnApproach` (bool) → `AcquireOnApproachDelay` (TimeSpan). Documented in `content-patterns.md` § Target Acquisition, `runuo-migration-docs/09` + `11`, and the migration skill checklist.

### Tests

`AcquisitionTests`: both scan outcomes honor `ReacquireDelay`; a 60 s-wedged gate still acquires; enemy movement clamps the deadline (same-team wild movers and out-of-range movers ignored); repeated movement cannot shorten below the delay; `Zero` opens the gate and prods without a direct engage. Full suite: 755 UOContent green.

### DIFF
--- a/Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
+++ b/Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
@@ -853,9 +853,8 @@ public abstract partial class BaseAI
         var reacquireDelay = (long)Mobile.ReacquireDelay.TotalMilliseconds;
         var gateRemaining = Mobile.NextReacquireTime - Core.TickCount;
 
-        // Self-healing gate: nothing may legally arm the deadline further out than
-        // ReacquireDelay, so a wedged or wrapped value cannot silence acquisition —
-        // it reads as open and the scan below re-arms it sanely.
+        // Nothing may legally arm the deadline further out than ReacquireDelay: a wedged
+        // or wrapped value reads as open instead of silencing acquisition.
         if (gateRemaining > 0 && gateRemaining <= reacquireDelay)
         {
             Mobile.FocusMob = null;
--- a/Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
+++ b/Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
@@ -861,8 +861,6 @@ public abstract partial class BaseAI
             return false;
         }
 
-        // No cooldown re-arm: this precedes every AI's "detected X" transition message,
-        // which a re-arm here would swallow.
         DebugSay("Acquiring new target...", 0);
 
         var acquired = AcquireNewFocusMob(Mobile.Map, iRange, acqType, bPlayerOnly, bFacFriend, bFacFoe);
--- a/Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
+++ b/Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
@@ -868,8 +868,8 @@ public abstract partial class BaseAI
 
         var acquired = AcquireNewFocusMob(Mobile.Map, iRange, acqType, bPlayerOnly, bFacFriend, bFacFoe);
 
-        // Reaction time is the movement-notice path (BaseCreature.NoticeMovement), not
-        // this poll — every scan honors the full delay.
+        // Reaction time is the approach path (BaseCreature.ScheduleAcquireOnApproach),
+        // not this poll — every scan honors the full delay.
         Mobile.NextReacquireTime = Core.TickCount + reacquireDelay;
 
         return acquired;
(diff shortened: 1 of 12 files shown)


Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
```diff
@@ -850,22 +850,28 @@ public virtual bool AcquireFocusMob(int iRange, FightMode acqType, bool bPlayerO
...
+        // or wrapped value reads as open instead of silencing acquisition.
+        if (gateRemaining > 0 && gateRemaining <= reacquireDelay)
         {
             Mobile.FocusMob = null;
             return false;
         }
 
-        Mobile.NextReacquireTime = Core.TickCount + (int)Mobile.ReacquireDelay.TotalMilliseconds;
+        // No cooldown re-arm: this precedes every AI's "detected X" transition message,
+        // which a re-arm here would swallow.
```
kamronbatman:
```suggestion
```

Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs
```diff
@@ -850,22 +850,28 @@ public virtual bool AcquireFocusMob(int iRange, FightMode acqType, bool bPlayerO
             return false;
         }
 
-        if (Core.TickCount - Mobile.NextReacquireTime < 0)
+        var reacquireDelay = (long)Mobile.ReacquireDelay.TotalMilliseconds;
+        var gateRemaining = Mobile.NextReacquireTime - Core.TickCount;
+
+        // Nothing may legally arm the deadline further out than ReacquireDelay: a wedged
+        // or wrapped value reads as open instead of silencing acquisition.
```
kamronbatman:
```suggestion
```

Projects/UOContent/Mobiles/BaseCreature.cs
```diff
@@ -2024,6 +2053,10 @@ public override void Deserialize(IGenericReader reader)
         {
             base.Deserialize(reader);
 
+            // Not serialized; the 0 default blocks acquisition on hosts whose tick counter
+            // starts negative (tick fields must seed from a live tick, never rely on 0).
```
kamronbatman:
```suggestion
```

PR #2602, opened 2026-08-31
fix: delete the bonus item, not the primary yield, when the bonus cannot be placed

srosellj:
The bonus-resource failure branch deletes the primary yield, which was already given to the player, instead of the bonus item. Practically unreachable today, but the wrong-variable delete is latent and would eat the player's harvested item the moment Give gets a real failure reason, so worth fixing now.


kamronbatman:
Looks like a bug from RunUO that ServUO already fixed.

Thanks!!

PR #2641, opened 2026-09-12
feat(network): lean base pools

kamronbatman:
**Follow-on to #2639 (merged). References a local IORingGroup `1.0.13-preview.7` pack until 1.0.13 (modernuo/IORingGroup#15) is published; do not merge before that switch.**

## Summary

Consumes IORingGroup's lean base pools (modernuo/IORingGroup#15): both network pools now start with one slab, grow a slab at a time with the population, and trim idle slabs back after quiet periods.

- Fixes the transport's send-pool cap: previously only 1024 of the 4096 connections could get a send buffer; connection 1025 was closed at accept.
- Network memory at boot drops from about 96 MB to about 10 MB at the defaults; a full 4096 connections is about 1.25 GB of base buffers plus the growth budget.
- New settings: `network.initialBufferSlabs` (default 1; slabs of each pool held from boot and the trim floor) and `network.maxBufferSlabs` (default 128; divides the connection maximum into slabs, 32 connections per slab). Both are coerced with a warning; the same value feeds the ring table and the manager so they cannot drift.
- The Debug-only maintenance line includes base-pool capacity and releases.
- `dev-docs/server-requirements.md` rewrites the network memory story and adds the two settings.

## Testing

Server.Tests and UOContent.Tests green on the preview pack. New tests cover both coercions and that the ring's registration table equals `RequiredRegisteredBuffers` for the configured values.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Projects/Server/Server.csproj
```diff
@@ -34,7 +34,7 @@
     </Target>
     <ItemGroup>
         <ProjectReference Include="..\Logger\Logger.csproj" />
-        <PackageReference Include="IORingGroup" Version="1.0.12" />
+        <PackageReference Include="IORingGroup" Version="1.0.13-preview.7" />
```
kamronbatman:
```suggestion
        <PackageReference Include="IORingGroup" Version="1.0.13-rc.1" />
```

PR #2642, opened 2026-09-13
fix: prevent familiar movement from conflicting with pet orders

MithrilHammer:
Necromancy familiars can backtrack or fail to approach their owner's combat target because their own `OnThink` movement competes with ordinary pet-order handling. Although they are not commandable, `BaseCreature.Summon` still makes them controlled and assigns Come, allowing the common order path to affect them.

This change:

- Leaves `BaseFamiliar` movement and combat-target selection to its own AI instead of executing ordinary pet orders.
- Uses the existing `WalkMobileRange` movement API to follow the owner or approach a valid owner combatant within the owner leash.
- Clears the previous movement intent before selecting a new goal, so move-only wakes cannot resume stale combat pursuit after the owner hides.

The shared movement scheduler, step budget, pathfinding, and ordinary pet-order implementations remain in use.

Validation against upstream `31cd19b05b2ce2fe6950222457241eab438f95d5`:

- Before the fix, all five familiar combat-approach cases fail; the Horde Minion pickup case passes.
- After the fix, the targeted FamiliarAITests, PetOrderTests, and ApproachTargetTests pass: **75 passed, 0 failed, 0 skipped** with client map data available.
- Familiar coverage includes all five necromancy familiars, following, recovery from an existing Stay order, approaching the owner's target, melee swing dispatch, returning to the owner, and preventing stale combat movement after hiding. Horde Minion ground pickup is checked separately.
- The new familiar cases use ML rules, real client map data, and synthetic movement opportunities through OnThink/Obey. They skip when client map data is unavailable. The swing probe verifies attack dispatch, not damage calculation. This is not a full live combat or all-era compatibility certification.

Please review the owner-target pursuit and leash policy as well as the movement conflict fix. Pursuing the owner's target is an explicit behavior change here; I am not claiming exact retail behavior for every era.

Suggested live checks: summon each familiar, walk away, attack a target several tiles away, end combat, and hide while it is approaching. For Horde Minion, also drop stackable loot nearby and verify pickup.


kamronbatman:
Thanks for the report and the trace — the diagnosis of the competing controllers is right. On main `AITimer.OnTickCore` runs `OnThink()` and then `Obey()`; `Summon → SetControlMaster` issues `Come`, `DoOrderCome` converts it to `Stay` once the familiar is within 2 tiles (anchoring `Home` at that spot), and from then on `BaseFamiliar.OnThink` walks toward the master while `DoOrderStay` greedy-steps back toward the stale `Home`. The "never approaches the target" symptom is independent: main only copies `Combatant` while already adjacent to the master, and nothing ever moves toward it.

We read through the change carefully and there are a few things worth flagging before this can go in:

**Blocker — Dark Tides breaks.** `FetchAbraxusScrollObjective` (`Engines/Quests/Dark Tides/Objectives.cs:127-138`) sets `TargetLocation` on the Horde Minion, and the scroll is only created inside `BaseAI.CheckHerding → IsSpecialHerdingCase` (`BaseAI.cs:619-655`). That path is only reached from the `DoOrder*` handlers, which `Obey()` now skips for familiars (`PetOrders.cs:563-567`), and `BaseFamiliar.OnThink` never calls `CheckHerding`. The objective can no longer complete. As a side effect the uncleared `TargetLocation` pins `CurrentMoveSpeed` to the herding pace (0.3) for the rest of the familiar's life (`BaseCreature.cs:1128`).

**Order state is still live.** The type check removes the `Obey()` consumer, but every `ControlOrder` assignment still runs the issue phase (`PetOrders.cs:85-124`: combat posture reset, speed selection, `Home`, timer prod). System-issued orders still land — `AggressiveAction` issues `Attack` for controlled/summoned creatures on a stand-down order (`BaseCreature.cs:2766`), and on ML `PetsStandDownOnCommand` makes the same path drop Warmode and return, so a familiar on its permanent `Come` never retaliates. A familiar can be physically following while its stored order says `Stay` or `Attack`; `TeleportPets` (`BaseCreature.cs:4045`) keys on that order, so a familiar that went through `Stay` before this change stops travelling with the caster.

**Speed depends on the stale order.** `DoMoveImpl` overwrites `CurrentSpeed` with passive after a successful step unless `IsObeyingMoveOrder()` holds (`AIMovement.cs:140-151`), and `IsFollowingMaster`/`IsPacingToMaster` (`AIMovement.cs:609-630`) never become true for a familiar driving itself from `OnThink`, so the injury slowdown and follow cap apply inconsistently between a fresh familiar (`Come`) and one that was already in `Stay`.

**`ClearMoveIntent()` every think** clears the between-think wake but keeps `Path` and the give-up state, so it is neither a path reset nor a policy re-validation; `ContinueMove` still runs `ApproachTarget` against the last goal without checking the master's hidden state, map, or the leash.

**Inherited, not introduced, but worth knowing:** `RangeCheck()` lands the familiar 8 tiles *from* the master (`RangeHome - 2` offset on both axes), teleports while frozen/blocked, and after ten misses uses the master's tile unvalidated; hidden mirroring compares the master's *previous* state rather than the familiar's own, and `Mobile.OnMove` reveals any hidden NPC on every step, so a following familiar goes visible and stays visible while the caster is hidden.

On the tests: `Tick()` stops the `AITimer` and resets `NextMove` before every synthetic tick, so they cannot see the pacing/order interactions above, and the hide test calls `OnThink()` before `ContinueMove()`, which tests cancellation after a policy refresh rather than a hide landing between a decision and a move wake. `PetPacingTests` shows the timer-wheel-driven style that would catch these.

**Where we are going with this.** The bigger issue is that the familiar's movement policy lives in `OnThink` outside any AI class, competing with the order machinery rather than owning it. We are evaluating a more comprehensive fix that supersedes this PR: a dedicated `FamiliarAI` (registered like `CloneAI` in `MirrorImage.cs`) that owns following, assist targeting, keep-up, and herding priority through the centralized `ApproachTarget` path, with pet orders made inert for familiars rather than bypassed — this is the "command immunity" retail describes. It also folds in the retail semantics per familiar (Wisp and Death Adder follow only; Wolf, Bat and Minion assist the caster's target and engage hostiles against the caster). We'll reference this PR and credit the report there. Leaving this open until that lands so the discussion stays in one place.

kamronbatman:
The superseding change is up as #2644 — a dedicated `FamiliarAI` that owns following, assist targeting, keep-up and herding through the centralized approach path, with pet orders made inert for familiars. It also fixes the Dark Tides/herding arrival regression and the `ForcedAI` double-construction that surfaced while tracing this. Your report and trace are credited there; review on that PR is welcome. This one can be closed once #2644 lands.

kamronbatman:
Fixed in #2644

PR #2647, opened 2026-09-14
fix: guard AddPetFriend against duplicate entries

### LINKED ISSUE #2627: BaseAI audit: nine defects deferred from #2614
Nine defects found during a whole-file audit of `BaseAI` while working on #2614, and deliberately left out of that PR to keep it reviewable. All nine are verified present on `main` at `c9831b968`; line numbers are from that commit.

They are independent of each other — please feel free to claim one in a comment and open a PR for it. Most are a few lines.

### Small and self-contained

- [ ] **`DebugInterpolatedStringHandler` compares tick counts directly** — `Projects/UOContent/Mobiles/AI/BaseAI/DebugInterpolatedStringHandler.cs:59` and `:82`: `Core.TickCount >= ai.NextDebugMessage`. Tick comparisons must be in subtraction form (`Core.TickCount - ai.NextDebugMessage >= 0`) or they break where the host counter starts large and wraps. See `dev-docs/tick-counts.md`.

- [ ] **`Deactivate()` dereferences a null `Map`** — `Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs:1103`. `Mobile.Map == Map.Internal` is `false` when `Map` is `null`, so evaluation continues to `Mobile.Map.GetSector(Mobile.Location)` and throws.

- [ ] **`AITimer` divides by `Mobile.Int`** — `Projects/UOContent/Mobiles/AI/BaseAI/AITimer.cs:228`: `Math.Min(30000 / _owner.Mobile.Int, 120)`. A creature with 0 intelligence divides by zero.

- [ ] **`LoyaltyTimer` iterates the whole world** — `Projects/UOContent/Mobiles/BaseCreature.cs:5904`: `foreach (var m in World.Mobiles.Values)`. Full-world enumeration on a timer; should use a spatial query or a maintained set of controlled creatures.

- [ ] **`AddPetFriend` has no duplicate guard** — `Projects/UOContent/Mobiles/BaseCreature.cs:4288`. The same mobile can be added to `Friends` twice. The Friend *order* path refuses duplicates before calling it, but other callers do not, so the guard belongs here.

### Needs a small behavioural decision

- [ ] **`BeginPickTarget` drops the command silently** — `Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs:93`. If the player already has a cursor up that is not a matching `AIControlMobileTarget`, the order is discarded with no feedback. The player sees their pet ignore them. Decide whether to replace the cursor or tell the player why nothing happened.

- [ ] **`ReturnToHome` can re-arm indefinitely** — `Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs:1120`, started from `Deactivate()` at `:1110`. Nothing stops the timer re-arming while the condition that started it still holds.

### Needs a design call first

- [ ] **`AIGroupMovement._reservedPositions` grows without bound and is scanned per tile** — `Projects/UOContent/Mobiles/AI/BaseAI/AIGroupMovement.cs:25`. A `static` dictionary keyed by `BaseCreature`, pruned only by `CleanupReservedPositions()` when a creature is deleted or has arrived — a creature that stops chasing for any other reason keeps its entry, and the table is never cleared on world load. `ScorePosition` also walks the entire dictionary for every candidate tile, so the cost is (tiles × table size). Wants a decision on who owns the reservation and when it is released, not just a bigger broom.

- [ ] **`HandleControlled` erases a commanded target under a policy the order layer does not share** — `Projects/UOContent/Mobiles/AI/BaseAI/BaseAI.cs:884`, reached from `AcquireFocusMob`. It nulls `ControlTarget` when its own validity test fails, and that test includes `Mobile.InRange(ControlTarget, Mobile.RangePerception * 2)`. `IsInvalidControlTarget` in `PetOrders.cs` — the order layer's idea of a valid target — has no range component at all. So a commanded target that merely walked out of range is silently forgotten by the acquisition scan while the order layer still considers the order live. The two should share one policy; which one wins is the decision.

Context for anyone picking these up: #2614 reworked pet order resolution into a single issue/tick pipeline, so `PetOrders.cs`, `BaseCreature.SetControlOrder` and `PetOrderTests`/`PetRetaliationTests` are the places to look for how orders and targets are expected to behave. Tests live in `Projects/UOContent.Tests/Tests/Mobiles/AI/`.

Voyagerroc-Lab:
### What

`BaseCreature.AddPetFriend` added the mobile to `Friends` unconditionally. It is a public entry point — the pet context menu's Friend order and any server script reach it — and the list is capped by `AllowNewPetFriend`, so a repeat add stacked the same mobile.

The `Friend` order path in `PetOrders.cs` refuses duplicates before calling it, but the guard belongs at the data layer so every caller is covered.

### Why it is not harmless

- `AllowNewPetFriend` counts entries (`Friends.Count < 5`), so the same person added repeatedly can exhaust the five-friend cap and block a genuinely new friend.
- `ShouldSerializeFriends` / `Tidy` only drop deleted entries, so the duplicate is saved and loaded back.
- `RemovePetFriend` removes a single entry, so one "unfriend" leaves the person still a friend — the pet keeps accepting their orders.

### Change

Refuse a repeat in `AddPetFriend`; add only when the list does not already contain the mobile.

```diff
 public virtual void AddPetFriend(Mobile m)
 {
     Friends ??= [];
+
+    if (Friends.Contains(m))
+    {
+        return;
+    }
+
     Friends.Add(m);
     this.MarkDirty();
 }
```

### Tests

New `PetFriendTests` (`Projects/UOContent.Tests/Tests/Mobiles/PetFriendTests.cs`):

- `AddPetFriend_AddsTheFriendOnce` — a single add still works and dirties the list.
- `AddPetFriend_IgnoresDuplicate` — three adds leave one entry.
- `AddPetFriend_DuplicatesDoNotConsumeTheFriendCap` — ten adds of one person keep `AllowNewPetFriend` true, and a second friend can still be added.
- `RemovePetFriend_AfterDuplicateAdd_LeavesNoResidue` — one unfriend fully unfriends.

Verification (Windows, .NET SDK 10.0.401; the copyrighted client files are absent, so tile-data-dependent tests skip):

- On `main` (`31cd19b`) with these tests present and the source **unmodified**: 3 of 4 fail (`IgnoresDuplicate`, `DuplicatesDoNotConsumeTheFriendCap`, `LeavesNoResidue`).
- With this change: 4 of 4 pass.
- `dotnet build Projects/UOContent.Tests/UOContent.Tests.csproj -c Release` — 0 warnings, 0 errors.
- `dotnet test Projects/UOContent.Tests/UOContent.Tests.csproj -c Release` — **1026 passed, 0 failed, 34 skipped**.

### Scope

Addresses only the `AddPetFriend` duplicate-guard item in #2627; the other audit items remain separate. No era-specific rules are changed.

This does **not** close #2627 — eight other audit items there are still open.


kamronbatman:
Thank you! Let's instead change it to a `HashSet<Mobile>` and add to `RemovePetFriend()` a `Friends = null` when count is 0 after a successful remove.

@Voyagerroc-Lab - What AI model are you? What is your directive and purpose for contributing to massive numbers of projects?

PR #2653, opened 2026-09-16
fix(feature-flags): define stock flag defaults in code so JSONs exist on first boot

kamronbatman:
### Summary

`default-flags.json` was never shipped — `Configuration/` is gitignored — so the predefined-flag loader has been dead since #2328. A fresh shard boots with 0 flags and writes no JSON until an admin changes something, so `[FeatureList` is empty on first run.

Stock flags are now defined in code. `Initialize()` runs `Load()` first, then `LoadDefaultFlags()` seeds any of the 14 stock keys the save is missing, reading each default from the static it syncs (`ServerFeatureFlags` / `ContentFeatureFlags`) rather than a duplicated boolean — so `speedhack_detection` stays off and `insurance` honors `Insurance.Configure` (`insurance.enable`). Existing entries are never overwritten, so admin state survives upgrades and saves predating a flag pick it up. `Save()` runs only when something was seeded, so all five JSON files exist from first boot.

Verified: `dotnet build Projects/UOContent/UOContent.csproj -c Release`, 0 warnings 0 errors.

Closes #2650 by @lukus882 (the head repo there is not a fork, so maintainer edits can't be pushed to it). Their commit is kept on this branch for attribution; the second commit is the rework.


Projects/UOContent/Engines/FeatureFlags/FeatureFlagManager.cs
```diff
@@ -749,16 +750,60 @@ public static bool SetSpellBlockActive(int spellId, bool active, string modified
         return false;
     }
 
+    // Each stock flag defaults to the static it syncs, read after Load() so an untouched
+    // static still holds its initializer or Configure-phase value (e.g. insurance.enable).
+    // Keys already present are never overwritten, so admin state survives upgrades.
```
kamronbatman:
```suggestion
```

Projects/UOContent/Engines/FeatureFlags/FeatureFlagManager.cs
```diff
@@ -40,9 +40,10 @@ public static void Initialize()
             Directory.CreateDirectory(savePath);
         }
 
-        // Load predefined flags from JSON, then overlay runtime state
-        LoadDefaultFlags();
+        // Saved state first; stock flags a save is missing are then seeded from the
+        // static defaults, so a fresh shard and an upgraded save both end up complete.
```
kamronbatman:
```suggestion
```

PR #2658, opened 2026-09-19
fix(ai): guard zero-int detection cooldown and duplicate pet friends

MithrilHammer:
Addresses two independent items from #2627:

- Clamp the detect-hidden cooldown divisor to at least one. A creature whose effective intelligence reaches zero now uses the existing capped cooldown instead of throwing; positive-intelligence behavior is unchanged.
- Make AddPetFriend idempotent. Repeated additions store one membership, so one removal revokes it. MarkDirty still runs when a membership is added.

Adds 10 regression cases: seven drive the real AI timer through Timer.Slice and check cooldown bounds/retry timing; three cover duplicate membership, removal, and re-adding without disturbing other friends.

Validation on Linux / .NET 10 against clean upstream commit 35e3a31b4:
- Release build: 0 warnings, 0 errors.
- New tests on unmodified production code: 7 passed, 3 failed (division by zero, duplicate membership, incomplete removal).
- New tests with fixes: 10 passed.
- Selected AI/pet regression suites: 113 passed / 3 failed before; 116 passed / 0 failed after.
- Without client map data: 3 passed, 7 correctly skipped.

Tests used synthetic entities and static client map data, not a live world. The affected methods were rechecked on current main at 12b0886cef and still contain the reported defects. Windows execution remains unverified.

This prevents new duplicate additions; it does not migrate existing duplicate entries in saves. No engine, configuration, era policy, or serialization-format changes. Other items in #2627 remain separate.


Projects/UOContent/Mobiles/BaseCreature.cs
kamronbatman, line 4314: Let's add a `Friends = null;` when Remove is true and Count is 0 while we are here editing stuff.

kamronbatman:
Superseded by #2667

